PR #5, opened 2017-07-23
Add corner IDs among other things

Adimote:
see the commit https://github.com/sourcebots/robot-api/commit/9c71836b68f1ef5912d1ef1c8565526237a40e93 for details


tests/mock_robotd.py
```diff
@@ -138,12 +143,21 @@ class MockCamera(Camera):
...
         super().__init__(node, camera)
 
 
+class MockGameState(GameState):
+    board_type_id = 'game'
+
+    def __init__(self, name, node):
+        super().__init__()
+        self._serial = name
+        _ = node  # Throw away node, not needed in this case
```
RealOrangeOne: what exactly is the point of this line? Except potentially silencing the linter

tests/mock_robotd.py
```diff
@@ -138,12 +143,21 @@ class MockCamera(Camera):
...
         self.serial = name
         # Create a camera with a file camera instead of a real camera
         super().__init__(node, camera)
 
 
+class MockGameState(GameState):
+    board_type_id = 'game'
+
+    def __init__(self, name, node):
+        super().__init__()
```
RealOrangeOne: the constructor for `GameState` takes parameters, you should be passing these in, or at least passing a stub

robot/board.py
```diff
@@ -50,23 +55,34 @@ def _send(self, message, retry=False):
...
-            if retry:
-                raise ConnectionError(self._get_lc_error())
-            else:
-                self._connect(self.sock_path)  # Reconnect
-                return self._recv(retry=True)  # Retry recursively
+
+        # TODO catch empty string receives as bad things
+        unended_text = b''
+
+        while not self._recv_buffer:
```
PeterJCLaw: This code block is very dense and seems rather overly complex.

I also think there's a bug in this logic. If we receive data which looks like: `b'foo\nbar'` then at line 82 we end up with `strings` being `[b'foo', b'bar']` and `unended_text` being `b'bar'`. Since that's a truthy value, we won't modify `_recv_buffer` and will loop around.

Whatever we next receive will correctly be appended to `unended_text`, though we'll have thrown away the contents of `strings`.

I think the logic we want looks something like this:
``` python
def receive():
    data = b''
    lines = []
    while True:
        data += receive_raw_from_socket_with_single_retry()

        if data:
            lines += data.split(b'\n')

            # An empty last line indicates the clean end of the data (why?)
            if lines[-1] == b'':
                return lines
            else:
                # Extract the incomplete last line as the prefix for the next
                # chunk of data we get
                data = lines.pop()
```

I think we can also tidy up the socket handling logic by adding a helper which looks something like:
``` python
def _socket_with_single_retry(self, handler)
    retryable_errors = (socket.timeout, BrokenPipeError, OSError)

    try:
        return handler(self.sock)
    except retryable_errors:
        pass

    # Retry once, need to reconnect first
    try:
        self._connect(self.sock_path)
    except FileNotFoundError:
        raise ConnectionError(self._get_lc_error())

    try:
        return handler(self.sock)
    except retryable_errors:
        raise ConnectionError(self._get_lc_error())
```

This can be used for all socket actions and then means that `receive_raw_from_socket_with_single_retry` is pretty simple (and could even be inlined):
``` python
def receive_raw_from_socket_with_single_retry():
    return self._socket_with_single_retry(
        lambda s: s.recv(Board.RECV_BUFFER),
    )
```

Thoughts?

robot/camera.py
```diff
@@ -48,7 +48,7 @@ def _cam_listener_worker(self):
 
         Works until `self._running` is set.
         """
-        while self._running.is_set():
+        while not self._stop.is_set():
```
PeterJCLaw: This isn't especially clear, particularly with the comment above referring to the old name. Maybe introduce a property `alive` (or some similar name) which is the used like:
``` python
while self.ailve:
    ...
```
The definition of the property can then be separately documented and hide the excessive negation inside it.

robot/power.py
```diff
@@ -21,12 +21,10 @@ def power_on(self):
         """
         Turn on power to all power board outputs
         """
-        self._send(json.dumps({'power': True}).encode('utf-8'))
-        self._recv()
+        self._send_recv(json.dumps({'power': True}).encode('utf-8'))
```
PeterJCLaw: This looks like we could use a `send_recv_data` method:
``` python
def send_recv_data(self, data):
    return json.loads(self._send_recv(json.dumps(data).encode('utf-8')))
```

tests/mock_robotd.py
Adimote: I can totally see how this is confusing, but in this case the parent is the robotd.GameState, I'm importing the actual robotd object to mock against, which takes no parameters as a constructor

README.md
```diff
@@ -4,11 +4,50 @@ This is the userspace API for the robot, which is what students will interface w
...
+# or
+markers = r.camera.see()
+```
+
+Servo positions can be set by:
 
 ``` python
-from robot import Robot
-robot = Robot() # TODO move this to the __init__.py so students don't need to do it
+r.servo_boards[0].ports[0].position = -1
```
Adimote: Sadly no, this isn't a problem right now as we're only giving them 1 of each board, but it's a good point which we shouldn't forget so I created an issue in robotd for it

README.md
```diff
@@ -4,11 +4,50 @@ This is the userspace API for the robot, which is what students will interface w
...
+# or
+r.servo_board.ports[0].position = -1
```
Adimote: it picks the first one, which sadly right now it's undefined as to which is first.

README.md
```diff
@@ -4,11 +4,50 @@ This is the userspace API for the robot, which is what students will interface w
...
+
+to run the tests, simply run `nosetests` (requires the `nose` python package) and it will run all tests
+
+
+# Zone ID script
+
+To run the zone script on USB insert, you must install the runusb script in [https://github.com/sourcebots/runusb]
+
+The zone ID script can be found at `robot/zone_script.py`, to use this for real robots you must put that script in a USB stick
+In the same directory as a file named `zone-<x>`, where `<x>` is the ID of the zone the robot should be with.
```
Adimote: Fixed this in 379e3d443ddc0419366b12834d1fd69bf4cfc0a2 then re-fixed it in db54a3be4575421a0fdae4affc013d62cf0d375e, and added full stops in 4ae00073df21a1270ba1bec19f360b07ffef7e0f

README.md
```diff
@@ -4,11 +4,50 @@ This is the userspace API for the robot, which is what students will interface w
...
+# or
+r.motor_board.m0.voltage = -1
 ```
+
+The zone the robot is going to start in can be gotten with:
+
+``` python
+r.zone
+# or
+r.game.zone
```
Adimote: Fixed in af0156f

robot/power.py
Adimote: fixed in f1c63b5

robot/board.py
```diff
@@ -50,23 +55,34 @@ def _send(self, message, retry=False):
...
+            strings = []
+            try:
+                output = self.sock.recv(Board.RECV_BUFFER)
+            except (socket.timeout, BrokenPipeError, OSError) as e:
+                if _retry:
+                    raise ConnectionError(self._get_lc_error())
+                else:
+                    print("Connection", self.sock_path, e, "retrying")
+                    self._connect(self.sock_path)
+                    return self._recv(_retry=True)  # Retry recursively
```
Adimote: Fixed in 6a3bfe7

robot/board.py
Adimote: I like it, implemented something similar but even more human readable, see 6a3bfe7

robot/board.py
```diff
@@ -2,20 +2,19 @@
 
 
 class Board:
-    SEND_TIMEOUT = 2
-    RECV_BUFFER = 4096*64
+    SEND_TIMEOUT = 20
+    RECV_BUFFER = 2048
```
Adimote: Fixed in 6a3bfe7 (should have split that commit into 2 in retrospect)

robot/board.py
```diff
@@ -50,23 +55,34 @@ def _send(self, message, retry=False):
                     raise ConnectionError(self._get_lc_error())
                 self._send(message, retry=True)  # Retry Recursively
 
-    def _recv(self, retry=False):
-
+    def _recv(self, _retry=False):
```
Adimote: Fixed in 6a3bfe7

robot/board.py
```diff
@@ -2,20 +2,19 @@
...
-        print("Receiving response:")
-        data = self._recv()
-        self._init_response(data)
-        print("Response:", data)
 
-    def _init_response(self, data):
-        pass  # Handle the response to the init command
+    def _greeting_response(self, data):
+        """Handle the response to the greeting command
+        NOTE: This is called on reconnect in addition to first connection"""
```
Adimote: Fixed in 423a829

README.md
```diff
@@ -4,11 +4,50 @@ This is the userspace API for the robot, which is what students will interface w
...
+```
+
+# Testing
+
+the robot-api tests require `robotd` (available at [https://github.com/sourcebots/robotd] ) in the python path (ie in the same directory as the `tests` folder ) Personally I create a soft link to the robotd directory
+
+to run the tests, simply run `nosetests` (requires the `nose` python package) and it will run all tests
```
Adimote: Fixed in c8b60d0

README.md
```diff
@@ -4,11 +4,50 @@ This is the userspace API for the robot, which is what students will interface w
...
+
+``` python
+r.zone
+# or
+r.game.zone
+```
+
+# Testing
+
+the robot-api tests require `robotd` (available at [https://github.com/sourcebots/robotd] ) in the python path (ie in the same directory as the `tests` folder ) Personally I create a soft link to the robotd directory
```
Adimote: Fixed in c8b60d0

robot/camera.py
Adimote: fixed in 8a03e37

robot/board.py
```diff
@@ -46,43 +47,53 @@ def _send(self, message, retry=False):
...
+        return self._socket_with_single_retry(
+            lambda s: s.recv(Board.RECV_BUFFER_BYTES),
+        )
+
+    def _recv(self, should_retry=True):
+        while b'\n' not in self.data:
+            if should_retry:
+                message = self.receive_raw_from_socket_with_single_retry()
+            else:
+                message = self.sock.recv()
```
PeterJCLaw: This call doesn't wrap failures in a `CommandError`. Is that intentional?

robot/board.py
```diff
@@ -46,43 +47,53 @@ def _send(self, message, retry=False):
...
+            if message is b'':
+                # Received blank, return blank
+                return b''
+
+            self.data += message
+        line = re.search(b'.*\n', self.data).group(0)
```
PeterJCLaw: You might enjoy the `maxsplit` argument to [`split`](https://docs.python.org/3/library/stdtypes.html#str.split) which would simplify this to:
``` python
line, self.data = self.data.split(b'\n', 1)
```

robot/board.py
```diff
@@ -81,15 +81,15 @@ def _socket_with_single_retry(self, handler):
...
+    def _receive_raw_from_socket_with_single_retry(self):
         return self._socket_with_single_retry(
             lambda s: s.recv(Board.RECV_BUFFER_BYTES),
         )
 
     def _recv(self, should_retry=True):
         while b'\n' not in self.data:
             if should_retry:
-                message = self.receive_raw_from_socket_with_single_retry()
+                message = self._receive_raw_from_socket_with_single_retry()
```
PeterJCLaw: Is there a reason not to inline this?

robot/board.py
```diff
@@ -46,43 +47,53 @@ def _send(self, message, retry=False):
...
-            if retry:
+            if _is_retry:
                 raise ConnectionError(self._get_lc_error())
             else:
                 try:
                     self._connect(self.sock_path)  # Reconnect
                 except FileNotFoundError:
                     raise ConnectionError(self._get_lc_error())
-                self._send(message, retry=True)  # Retry Recursively
+                self._send(message, _is_retry=True)  # Retry Recursively
```
PeterJCLaw: Why not use the `_socket_with_single_retry` helper in this method too?

robot/robot.py
```diff
@@ -89,12 +89,75 @@ def servo_boards(self):
...
+        boards = self._update_boards(self.known_gamestates, GameState, 'game')
+        self.known_gamestates = boards
+        return self._dictify_boards(boards)
+
+    @staticmethod
+    def _single_index(name, list_of_boards):
+        if list_of_boards:
+            return list_of_boards[0]
+        else:
+            raise ValueError("No {}s connected".format(name))
```
PeterJCLaw: Minor: As this is raised during property access, is `AttributeError` more usual?

robot/board.py
```diff
@@ -41,32 +53,53 @@ def _send(self, message, retry=False):
         try:
             self.sock.send(message)
         except (socket.timeout, BrokenPipeError, OSError):
-            if retry:
+            if _is_retry:
```
PeterJCLaw: I probably should have asked earlier: what's the purpose of the retries? Is it because we think the connection is flaky or is it a result of using single-message sockets?

If the latter, I'm not convinced that retries are the best way to go. I wonder whether explicitly setting up and tearing down the socket would be better in this case.

Adimote: It's to avoid for flaky connections

robot/board.py
Adimote: Fixed in d1ae187caa177e40d3569a15fce41c46d3ea8673

robot/board.py
Adimote: Good point, fixed in 5418901da748519558b179a73aa1b06319e0c1dc

robot/board.py
Adimote: Fixed in 62cf5c031f8f567b9ebdb0fe0500251492322b4b

robot/board.py
```diff
@@ -40,6 +41,9 @@ def _send_recv(self, message):
         self._send(message)
         return self._recv()
 
+    def _send_recv_data(self, data):
+            return json.loads(self._send_recv(json.dumps(data).encode('utf-8')))
```
Adimote: Yes, it either returns a json string or an empty string.

PeterJCLaw: Hrm, I suspect we'll get errors some of the time then:
``` python
>>> print(json.loads(b''))
  File "python", line 3, in <module>
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

robot/board.py
PeterJCLaw: Apologies, I'd meant inlining the body of the awkwardly named function (rather than moving to a ternary-style statement)

robot/board.py
```diff
@@ -81,8 +81,8 @@ def _recv(self, should_retry=True):
                 return b''
 
             self.data += message
-        line = self.data.split(b'\n', 1)
-        self.data = self.data[len(line):]
+        line = self.data.split(b'\n', 1)[0]
```
PeterJCLaw: Maybe add a comment here about why we're not using unpacking for this (which feels like the more natural solution)?

(I also wondered about using `*args`, but I'm not sure doing so ends up any clearer than what's already here)

PR #6, opened 2017-07-23
Create BoardList structure

RealOrangeOne:
Fixes https://github.com/sourcebots/robot-api/blob/master/robot/robot.py#L53

Allow boards to be indexed by serial number and integer index. Also prevent modification to the boards list

### DIFF
--- a/robot/board.py
+++ b/robot/board.py
@@ -1,6 +1,19 @@
 import socket
 
 
+class BoardList(dict):
+    def __getitem__(self, attr):
+        if type(attr) is int:
+            return super().__getitem__(list(self.keys())[attr])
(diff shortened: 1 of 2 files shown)


Adimote:
Actually there were a few issues with the PR. it didn't implement `__contains__` so 

    0 in r.motor_boards
    >>> False # should be True

also the line 

    return BoardList({board.serial for board in boards})

should have been 

    return BoardList({board.serial: board for board in boards})

Adimote:
Also I pushed in 1169bb1943769d696511f1b4f215311aff0c37d6 using an abstract base class for the dictionary, as it can magically give us bonus features (such as __contains__)

PR #9, opened 2017-07-28
add Circle CI

Adimote:


Adimote:
Ignoring the fact that the circleci passes are failing on the PR which adds Circleci checks! This isn't ready to merge because the checks are passing, but not halting.

PR #20, opened 2017-08-02
Ignore connection failures for the boards

thomasleese:


robot/robot.py
```diff
@@ -58,7 +58,12 @@ def _update_boards(self, known_boards, board_type, directory_name):
         # Add all boards that weren't previously there
         for board in new_paths - known_paths:
             print('New board found:', board)
-            boards.append(board_type(board))
+
+            try:
+                new_board = board_type(board)
+                boards.append(new_board)
+            except ConnectionRefusedError as e:
```
Adimote: I'd also catch `BrokenPipeError`, `OSError`, `ConnectionResetError`, and `FileNotFoundError` just to be safe, as those are caught when retrying

thomasleese: `FileNotFoundError` I would agree, the others shouldn't happen during a call to `socket.connect`.

PR #21, opened 2017-08-02
Fix tests

Adimote:
currently the tests are failing, this fixes that


tests/test_camera.py
```diff
@@ -1,76 +0,0 @@
-import os
```
RealOrangeOne: Did these camera tests still pass? If so i'd suggest leaving them in, unless you want to remove _all_ the camera api with it

Adimote: they didn't pass

robot/robot.py
```diff
@@ -167,7 +174,14 @@ def mode(self):
...
         return self._game.mode
 
-    def __del__(self):
+    def close(self):
         # stop the polling threads
         for camera in self.known_cameras:
             camera._stop_poll()
+        for board_type in self.all_known_boards:
+            for board in board_type:
+                board._clean_up()
```
RealOrangeOne: I'd suggest using `del board` here. So custom deconstructors can be used on each board if necessary

RealOrangeOne:
Use images from `sb-vision`. The image locations are known, and will provide a better and more standard test

PR #22, opened 2017-11-04
Move serial number methods to base board class

RealOrangeOne:
Rather than always re-defining the same serial logic in each subclass of `Board`, move it up into `Board`.


robot/board.py
```diff
@@ -41,6 +42,14 @@ def __init__(self, socket_path):
         self.data = b''
 
         self._connect(socket_path)
+        self._serial = Path(socket_path).stem
+
+    @property
+    def serial(self):
```
thomasleese: Since this property does nothing but expose the `_serial` variable, I would just make the variable a public attribute and document is in the docstring for the class (as per [PEP 257](https://www.python.org/dev/peps/pep-0257/#id17)).

robot/board.py
```diff
@@ -41,6 +42,14 @@ def __init__(self, socket_path):
         self.data = b''
 
         self._connect(socket_path)
+        self._serial = Path(socket_path).stem
```
thomasleese: Perhaps not relevant for this PR, but I would maybe wrap the `self.socket_path` as a `Path` which then may be of use further down the class. I notice in the `_connect` method we seem to be passing a `socket_path` but then the method just goes on to use the `self.socket_path` and then `str`s it, so it being a `Path` wouldn't be an issue.

robot/board.py
RealOrangeOne: The main benefit of storing it as an internal variable is it's immutable, and means there's little chance of them accidentally overriding it.

thomasleese: Hmm, yeah, that seems reasonable enough. I would be tempted then to make the `serial` property just call `self.socket_path.stem` and not store it as a `_serial` attribute. If that's slow (which I don't think it is) there is always [@cached_property](https://github.com/pydanny/cached-property).

RealOrangeOne: I literally just made that change ;)

kierdavis:
Looks good to me. Are the tests failing because of this PR or are they currently broken?

RealOrangeOne:
Looks like it's because theyre broken? cc @Adimote

kierdavis:
Although this PR's diff looks good to me, the tests are failing (and not on master), so I'd prefer we didn't merge this until this is fixed.

Having said that, I can't for the life of me figure out why they're failing. They seem to be fine in 80f190f (more or less, there are [two failures out of sixteen][1]), which is the first commit in this PR; however in the immediately following commit, e520f04, [every single test produces an error][2].

After a bit of investigation, my best guess as to what is causing the error scenario is:
* `_connect` calls `receive`
* `receive` calls `_recv_from_socket` using the `_socket_with_single_retry` wrapper
* `_recv_from_socket` times out
* In response, `_socket_with_single_retry` attempts to reconnect the socket and retry the operation
* Reconnecting the socket involves calling `_connect`
* As a result, everything recursively repeats from the beginning
* After about 6 consecutive failures we run out of some kind of resource (open sockets?)
* `socket.connect` thus fails with a "Resource temporarily unavailable" exception

Perhaps this could be fixed by not allowing the `receive` to retry when it is called by `_connect`? `receive` appears to already have a parameter `should_retry`, so this should be an easy fix (if it is indeed the right way to approach this).

However I have no clue how (if?) this is caused by the changes in e520f04. After some trial-and-error I've found that applying [this change][3] on top of this PR doesn't result in the error scenario above being triggered, and [all tests pass][4] (though the linting then fails). This change constitutes going back to using a `str` for `socket_path` rather than a `Path`. 100 brownie points will be awarded to whoever can explain why this patch works.

tl;dr: it seems wrapping `socket_path` in a `Path` somehow causes socket reads to time out and results in all tests failing

[1]: https://circleci.com/gh/sourcebots/robot-api/285
[2]: https://circleci.com/gh/sourcebots/robot-api/286
[3]: https://github.com/sourcebots/robot-api/commit/04c71fe41a765a3b28c909f3bfeabd2ec1cb46d3
[4]: https://circleci.com/gh/sourcebots/robot-api/288

robot/servo.py
```diff
@@ -114,13 +112,6 @@ def __init__(self, socket_path):
             for x in gpio_pins
         }
 
-    @property
-    def serial(self):
-        """Serial number of the board."""
-        return self._serial
-
-    # Servo code
```
PeterJCLaw: Unless they're wrong, I'd be in favour of keeping the comments which group the methods into "Servo" and "GPIO" chunks.

robot/board.py
```diff
@@ -56,7 +63,7 @@ def _connect(self, socket_path):
         try:
             self.socket.connect(str(self.socket_path))
         except ConnectionRefusedError as e:
-            print('Error connecting to:', socket_path)
+            print('Error connecting to: ', self.socket_path)
```
PeterJCLaw: I don't think we need the extra space here (though the change to add `self.` on `socket_path` is correct):
``` python
>>> print('a', 42)
a 42
>>> print('a ', 42)
a  42
```

RealOrangeOne: hmm, odd, i'd always assumed `print` just did a concatenation behind the scenes. Will adjust

PR #25, opened 2017-11-05
Add camera tests

Adimote:
Will only pass tests once https://github.com/sourcebots/robotd/pull/24 is merged, re-adds the camera tests


robot/camera.py
```diff
@@ -76,7 +73,7 @@ def _see_to_results(data):
         for token in data["markers"]:
             markers.append(Marker(token))
         # Sort by distance
-        return ResultList(sorted(markers, key=lambda x: x.distance_metres))
+        return ResultList(markers)
```
RealOrangeOne: I would still argue this list should be sorted by distance. Unless it's handled elsewhere?

robot/robot.py
```diff
@@ -167,7 +174,14 @@ def mode(self):
...
         return self._game.mode
 
-    def __del__(self):
+    def close(self):
         # stop the polling threads
         for camera in self.known_cameras:
             camera._stop_poll()
+        for board_type in self.all_known_boards:
+            for board in board_type:
+                board._clean_up()
```
RealOrangeOne: I'd suggest renaming `_clean_up` to use the `__del__` method, so it's automatically called when a board is destroyed, and means we're not adding additional custom functionality

robot/camera.py
RealOrangeOne: I would still argue this list should be sorted by distance. Unless it's handled elsewhere?

Adimote: without support of Polar co-ordinates it's not trivial to sort by distance any more. Feature coming once https://github.com/sourcebots/sb-vision/pull/5 is merged

robot/robot.py
thomasleese: I agree, although probably worth also doing an explicit `del board` too.

PR #26, opened 2017-11-05
Add polar coords

Adimote:
Will only pass tests once https://github.com/sourcebots/robotd/pull/24 and https://github.com/sourcebots/sb-vision/pull/5 are merged.


Adimote:
It would be, wouldn't it?

PR #27, opened 2017-11-07
Initialization Updates

RealOrangeOne:
- Poll for start button more frequently
- Add logging messages to setup


RealOrangeOne:
@Adimote not exactly sure what's up with the tests here?

robot/robot.py
```diff
@@ -25,30 +25,48 @@ def __init__(self, robotd_path=ROBOTD_ADDRESS):
...
         self.known_gamestates = []
+        print("Initializing Hardware...")
         self.all_known_boards = [
             self.known_power_boards,
             self.known_motor_boards,
             self.known_servo_boards,
             self.known_cameras,
             self.known_gamestates,
         ]
+        self.physical_boards = [
```
PeterJCLaw: Unless the "serial" of the `Game` is particularly undesirable (I don't know what it is), I'd encourage just including it in the output. Having _another_ list of lists feels overkill for this functionality.

RealOrangeOne: It's not undesirable, I just couldnt think of a good serial number, as the current implementation lacks one. Perhaps just a fixed string would work

robot/board.py
```diff
@@ -168,4 +168,10 @@ def close(self):
         """
         self.socket.detach()
 
+    def __str__(self):
+        return "{} - {}".format(
+            self.__name__,
+            self.serial
```
PeterJCLaw: Missing trailing comma here (though I actually think this would fit on a single line?)

robot/robot.py
```diff
@@ -37,22 +38,33 @@ def __init__(self, robotd_path=ROBOTD_ADDRESS):
...
         self.power_board.power_on()
+        print("Starting user code.")
+
+    def _display_boards(self):
+        print("Found the following hardware devices:")
+        for board_list in self.all_known_boards:
+            for board in board_list:
+                print(str(board))  # Force string representation
 
+    def wait_start(self):
```
PeterJCLaw: I wonder if we should push this down into the `PowerBoard` class?

robot/robot.py
PeterJCLaw: My reading of the code suggests that `GameState` does indeed have a `.serial`, though it'll end up being the constant value `'state'`.

robot/robot.py
RealOrangeOne: Hmm, interesting idea, i'll see how easy it is to migrate the code

robot/robot.py
```diff
@@ -38,23 +38,26 @@ def __init__(self, robotd_path=ROBOTD_ADDRESS):
...
+        self.power_board.wait_start()
+
+        self.power_board.power_on()
+        print("Starting user code.")
+
+    def _display_boards(self):
+        print("Found the following hardware devices:")
+        for board_list in self.all_known_boards:
+            for board in board_list:
+                print(str(board))  # Force string representation
```
PeterJCLaw: Isn't `str(x)` the default behaviour for `print(x)`? (genuinely curious as I'd assumed it was but I don't think I've actually seen any docs on it)

robot/robot.py
```diff
@@ -38,23 +38,26 @@ def __init__(self, robotd_path=ROBOTD_ADDRESS):
             self.known_gamestates,
         ]
 
+        self._display_boards()
```
PeterJCLaw: I'm unfortunately fairly sure that the boards lists will be empty at this point. This is because we currently only initialise them lazily, something I'd not realised until I ran the full stack locally this week.

We could force the initialisation of the devices here, though that then leads to the wondering of whether the boards themselves should be logging their initialisation, rather than it being a global thing. If we had proper logging (waaay out of scope for this PR), I'd probably suggest doing both, under different loggers, with the logger here outputting to the main log and keeping the per-board messages for debugging.

robot/robot.py
RealOrangeOne: Something tells me it uses `__repr__` by default here, or something weird like that. Either way, better explicit

PeterJCLaw: Given the surprise factor of this being explicit, I actually think this is better implicit. My reasoning is that, as it stands, this leads to readers questioning what would otherwise happen.  (I'd draw similarities with checking for `None` where it wouldn't reasonably be expected, etc.)

Anyway, it look like `str` _is_ the default behaviour:
``` python
>>> class Foo:
...  def __str__(self):
...   return 'Foo.__str__'
...
>>> repr(Foo())
'<__main__.Foo object at 0x000001AAF0379780>'
>>> str(Foo())
'Foo.__str__'
>>> print(Foo())
Foo.__str__
```

Recall however that the default `__str__` calls `__repr__`.

robot/robot.py
kierdavis: Agreed totally. I think this line (and therefore the `_display_boards` function) can be removed without issue - we still log when boards are connected to for the first time ("New board found: /var/robotd/power/xxx").

robot/robot.py
```diff
@@ -38,23 +38,26 @@ def __init__(self, robotd_path=ROBOTD_ADDRESS):
...
         self._wait_for_power_board()
 
+        print('Waiting for start button.')
+        self.power_board.wait_start()
+
+        self.power_board.power_on()
```
kierdavis: This should be done before `wait_start` rather than after, to give boards (e.g. motor board) time to initialise and report themselves ready before the user code starts trying to use them (which may be immediately after `__init__` returns).

robot/robot.py
```diff
@@ -39,22 +40,18 @@ def __init__(self, robotd_path: Union[str, Path]=ROBOTD_ADDRESS) -> None:
...
 
         self._wait_for_power_board()
+        self.power_board.power_on()
+
+        print('Waiting for start button.')
+        self.power_board.wait_start()
+
+        print("Starting user code.")
 
     def _wait_for_power_board(self):
```
PeterJCLaw: Can we rename this method to make sense? It no longer does any waiting, and certainly not for the power board.

tests/test_power.py
```diff
@@ -14,21 +14,13 @@ def setUp(self):
         self.robot = Robot(robotd_path="/tmp/robotd")
 
     def test_on_off(self):
+        self.power_board.message_queue.get()
```
kierdavis: This isn't needed now that we've gone back the original order of having the outputs come on before the start LED, and it's causing the tests to fail. Intuition suggests you might also need a `clear_queue` between lines 22 and 24 to throw away any start LED stuff, but I'm not sure.

kierdavis: I've fixed this in #66.

PR #29, opened 2017-11-19
Add debian package

Adimote:
Make it a debian package so we can ship it straight to the robots instead of as an interface in the USB sticks


debian/rules
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/make -f
+
+export PYBUILD_NAME=robot-api
+export PYBUILD_RULES=3.4
+
+%:
+	dh $@ --with python3 --buildsystem=pybuild
+
+override_dh_install:
+	find debian/ -name __pycache__ -type d | xargs -r rm -r
```
PeterJCLaw: You might enjoy the `-delete` option to `find`

README.md
```diff
@@ -2,6 +2,35 @@
...
+debuild -uc -us
+
+```
+
+## Instaling the debian package
+
+After building, the .deb file should appear in the parent directory, just run
+
+``` bash
+sudo dpkg -i
```
PeterJCLaw: Does this not need the name of the package file to be passed on the command line?

debian/rules
PeterJCLaw: (Is there not a better way of excluding these files from being included?)

README.md
```diff
@@ -2,6 +2,38 @@
...
+
+# Install dependencies
+sudo mk-build-deps -ir
+
+# Build the package:
+debuild -uc -us
+
+```
+
+## Instaling the debian package
```
prophile: "Installing"

Adimote: Huh, the commit message was also wrong, it appears I don't know how to spell intsall.

PR #31, opened 2017-11-21
Buzz args

RealOrangeOne:
- Force buzz to only take `note` or `frequency`
- Add tests 🎉 

(Docs will need updating after this)


robot/power.py
```diff
@@ -54,8 +54,8 @@ def start_button_pressed(self):
         return status["start-button"]
 
     def buzz(self, duration, note=None, frequency=None):
-        if note is None and frequency is None:
-            raise ValueError("Either note or frequency must be provided")
+        if bool(note) == bool(frequency):
+            raise ValueError("Please only provide note or frequency")
```
PeterJCLaw: This allows the caller to use this API as the following, which is likely confusing and silly. I'm also not convinced that this check is as clear as it could be (though it is clever).

``` python
board.buzz(4, False, 4)
```

Note that  this will also then fail since the check `note is not None` will be truthy and the user will get the possibly confusing `KeyError("False is an invalid note")`.

I suggest going back to the previous check (though maybe that could raise a `TypeError` instead?) and adding a separate check that at least one is provided.

tests/test_power.py
```diff
@@ -62,5 +62,33 @@ def test_remove_board(self):
...
         self.assertTrue('ABC' not in self.robot.power_boards)
 
+    def test_buzz_args(self):
+        with self.assertRaises(ValueError):
+            self.robot.power_board.buzz(1, note='c', frequency=100)
+
+        with self.assertRaises(ValueError):
+            self.robot.power_board.buzz(1)
+
+        with self.assertRaises(KeyError):
```
PeterJCLaw: i'd be tempted to break the note validity checks out from the argument presence checks.

I'd also suggest adding tests which explore whether the case of the note matters.

robot/power.py
```diff
@@ -53,13 +53,13 @@ def start_button_pressed(self):
...
             raise ValueError("Either note or frequency must be provided")
+        if note is not None and frequency is not None:
+            raise ValueError("Only provide note or frequency")
         if note is not None:
-            if note not in self.BUZZ_NOTES:
-                raise KeyError("{} is an invalid note".format(note))
-            frequency = self.BUZZ_NOTES.get(note.lower())
+            frequency = self.BUZZ_NOTES[note.lower()]
         if frequency is None:
             raise ValueError("Invalid frequency")
```
PeterJCLaw: While we're tidying the arguments it might be nice to include the actual value of `frequency` in this error message.

RealOrangeOne:
Yes, the `KeyError` is raised with the invalid key in the exception message, which is the value of `note.lower()`

PR #33, opened 2017-11-26
Fix package requirements

kierdavis:
For some reason the `requirements_test.txt` file listed `pycparser` as a requirement, although its never used anywhere. I presume this was left over from an earlier time?

Also, the tests require `robotd` and `sb-vision` to be available.

I've unfamiliar with python packaging idioms - have I done this correctly?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,7 @@ setup(
     author='SourceBots',
     license='MIT',
     install_requires=[],
-    dependency_links=['git+ssh://git@github.com/sourcebots/robotd.git#egg=robotd'],
-    tests_require=[],
+    dependency_links=[],
+    tests_require=["robotd", "sb-vision"],
     packages=find_packages(exclude=['contrib', 'docs', 'tests']),
 )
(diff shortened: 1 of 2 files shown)


kierdavis:
Oh, it looks like adding `robotd` and `sb-vision` to `requirements_test.txt` breaks the CI, since they're installed separately. I guess I can add them to `setup.py` but not to `requirements_test.txt`?

setup.py
```diff
@@ -9,6 +9,6 @@
     license='MIT',
     install_requires=[],
     dependency_links=['git+ssh://git@github.com/sourcebots/robotd.git#egg=robotd'],
```
RealOrangeOne: Not _entirely_ related to this PR, but seeing as `robotd` isnt an explicit requirement of using `robot-api`, we should probably remove it from `dependency_links` too

kierdavis: Good point, and I would consider it relevant to this PR. You don't need to have `robotd` installed in order to use `robot-api`; you just need something that's compatible with `robotd`'s API.

PR #34, opened 2017-12-01
Type checking!

RealOrangeOne:
Add type checking, powered by MyPy 🎉 

- [x] Merge #22, as currently the base `Board` class doesn't implement `serial`
- [x] Work out a fix for `robot/camera.py:20: error: Signature of "__getitem__" incompatible with supertype "list"` #54


RealOrangeOne:
Turns out the syntax i'm using for types is only supported in python3.6+, will refactor to use comment-based types instead

.circleci/config.yml
```diff
@@ -51,6 +51,9 @@ jobs:
       - run:
           name: run tests
           command: python3 -m unittest
+      - run:
+          name: type check
+          command: mypy --strict-optional --ignore-missing-imports --follow-imports=skip robot/
```
PeterJCLaw: Can we set these options in a config file somewhere so editors can also pick them up? (and/or so that just running `mypy` without specific arguments, which is what I expect editors will do, will pick them up)

robot/camera.py
```diff
@@ -61,7 +104,7 @@ def _cam_listener_worker(self):
                     self._latest = data
 
     @staticmethod
-    def _see_to_results(data):
+    def _see_to_results(data) -> ResultList:
```
PeterJCLaw: Please can we also type `data`

robot/camera.py
```diff
@@ -6,6 +6,49 @@
 from robot.markers import Marker
 
 
+class ResultList(MutableSequence):
```
PeterJCLaw: Why does this definition need to move?

robot/game.py
```diff
@@ -36,12 +36,10 @@ def zone(self):
...
+    def mode(self) -> GameMode:
         """
         Whether or not the robot is in competition mode
         :return: if the robot is in competition mode
         """
         value = self.send_and_receive({})['mode']
-        for enum in GameMode:
-            if value == enum.value:
-                return enum
+        return GameMode(value)
```
PeterJCLaw: Please can we _not_ edit functionality (even when it's equivalent) as part of this PR? (The easiest way to do this is to cherry-pick the fixes onto master separately and then merge master outwards into this branch) This makes it clearer what changes are being made.

Alternatively, please can we make it clear in the PR (and merge message) that this will also fix issues essential to making the types work.

robot/markers.py
```diff
@@ -1,32 +1,10 @@
...
-    @property
-    def y(self):
-        """ Y co-ordinate of the cartesian position"""
-        return self._y
-
-    @property
-    def z(self):
-        """ Z co-ordinate of the cartesian position"""
-        return self._z
+CartCoord = namedtuple("CardCoord", ["x", "y", "z"])
```
PeterJCLaw: See comment about not changing functionality in this PR.

robot/markers.py
```diff
@@ -92,20 +70,15 @@ class Marker:
     def __init__(self, data):
         self._raw_data = data
 
-        # Go through all the data, add an _ at the start.
-        data = {"_" + k: v for k, v in data.items()}
-
-        self.__dict__.update(data)
-
```
PeterJCLaw: ewww.

robot/markers.py
```diff
@@ -92,20 +70,15 @@ class Marker:
...
     @property
-    def id(self):
+    def id(self) -> int:
         """ID of the marker seen"""
-        return self._id
+        return self._raw_data['id']
 
     @property
-    def size(self):
+    def size(self) -> Tuple:
```
PeterJCLaw: What's this a tuple of? I'd expect to see `Tuple[int, str]` or similar.

robot/markers.py
```diff
@@ -117,24 +90,23 @@ def size(self):
...
 
-    def is_token_marker(self):
+    def is_token_marker(self) -> bool:
         """ If the marker is a token marker """
         return self.id in TOKEN
 
     @property
     def cartesian(self):
         raise NotImplementedError("This is not implemented.")
-        return CartCoord()
```
PeterJCLaw: Why not convert this to a return type annotation?
Do we need to change this to being marked as an abstract property?

robot/robot.py
```diff
@@ -18,13 +21,13 @@ class Robot:
 
     ROBOTD_ADDRESS = "/var/robotd"
 
-    def __init__(self, robotd_path=ROBOTD_ADDRESS):
+    def __init__(self, robotd_path:str=ROBOTD_ADDRESS) -> None:
```
PeterJCLaw: I think this wants a space before the `str` and after the colon: `robotd_path: str=...`.

robot/markers.py
PeterJCLaw: You might enjoy the `typing.NamedTuple`.

.circleci/config.yml
prophile: These can be configured in `setup.ini` I think.

robot/markers.py
```diff
@@ -92,20 +70,15 @@ class Marker:
...
         self._raw_data = data
 
-        # Go through all the data, add an _ at the start.
-        data = {"_" + k: v for k, v in data.items()}
-
-        self.__dict__.update(data)
-
     @property
-    def id(self):
+    def id(self) -> int:
```
prophile: I would greatly enjoy using `NewType` for this sort of thing.

robot/robot.py
```diff
@@ -161,7 +164,7 @@ def _game(self):
         return self._single_index("game states", self._games)
 
     @property
-    def zone(self):
+    def zone(self) -> int:
```
prophile: Here again I would enjoy a `NewType`.

robot/servo.py
```diff
@@ -141,10 +141,10 @@ def _get_pin_mode(self, pin):
     def _set_pin_mode(self, pin, value: PinMode):
         self.send_and_receive({'pins': {pin: value.value}})
 
-    def read_analogue(self):
+    def read_analogue(self) -> dict:
```
prophile: Can we put a more precise type here?

robot/servo.py
```diff
@@ -141,10 +141,10 @@ def _get_pin_mode(self, pin):
...
         command = {'read-analogue': True}
         return self.send_and_receive(command)['analogue-values']
 
-    def read_ultrasound(self, trigger_pin, echo_pin):
+    def read_ultrasound(self, trigger_pin, echo_pin) -> float:
```
prophile: Can we type the pin arguments?

robot/robot.py
```diff
@@ -1,13 +1,16 @@
...
 
 from robot.board import BoardList
 from robot.camera import Camera
-from robot.game import GameState
+from robot.game import GameState, GameMode
 from robot.motor import MotorBoard
 from robot.power import PowerBoard
 from robot.servo import ServoBoard
 
+Board = Union[Camera, GameState, MotorBoard, PowerBoard, ServoBoard]
```
prophile: Do we semantically want this rather than using the base class? I.e. even in the presence of new subclasses, `Board` is closed to these five specifically?

robot/markers.py
```diff
@@ -76,7 +54,7 @@ def rot_y_deg(self):
         return math.degrees(self._rot_y_rad)
 
     @property
-    def distance_metres(self):
+    def distance_metres(self) -> float:
```
prophile: If we were going full type system nuttery I'd advocate for having this return `Metres` as a newtype of floats. Ditto for degrees above.

robot/camera.py
prophile: We could also take the opportunity to un-break this class since it currently won't typecheck properly anyway.

PeterJCLaw:
Note: this PR is likely to conflict heavily with #41. Since this PR already has some conflicts, I suggest we merge that one first and then rework this PR.

I'm also tempted to use `flake8-mypy` to pull the type checking into the linting, though I've not explored whether that actually works before.

robot/robot.py
RealOrangeOne: I originally used the base class, but for some reason it didn't like it. I should certainly have another play at some point

.circleci/config.yml
PeterJCLaw: Indeed, `mypy` reads `setup.cfg`. I've been putting all our lint configuration there as it has the benefits of being a single easy-to-see place where they're all defined as well as allowing for trivial invocations of all the tools. The latter is useful for editor support too.

See also https://github.com/sourcebots/sb-vision/pull/21.

robot/camera.py
RealOrangeOne: I think building an accurate type-based schema for the results of `robotd` would be great, but maybe OOS for this PR. Doing something like `Dict[str, Any]` is accurate, but not especially useful. 

I'm not exactly sure if you can specify keys for dictionaries yet, I have a slight feeling you can't? If not then perhaps I can just use a standard `JSON` type, from https://github.com/python/typing/issues/182

robot/robot.py
RealOrangeOne: As does the linter!

robot/game.py
```diff
@@ -3,7 +3,7 @@
 from typing import NewType
 from robot.board import Board
 
-Zone = NewType('Zone', int)
+TZone = NewType('TZone', int)
```
PeterJCLaw: Ah, apologies. My comment about naming types like this was intended to refer only to generic type variable types -- since the type isn't concrete there's a desire to show that.

For concrete types (even when they're newly defined types), I think we should stick with just having a plain type name.

robot/camera.py
```diff
@@ -1,9 +1,30 @@
...
+    """
+
+    def __getitem__(self, index: int) -> Marker:
+        try:
+            return super().__getitem__(index)
+        except IndexError as e:
+            if not self:
+                raise IndexError("Trying to index an empty list") from None
+            else:
+                raise e
```
PeterJCLaw: minor: this should probably be `raise` rather than `raise e`

robot/markers.py
```diff
@@ -1,30 +1,15 @@
...
-    @property
-    def y(self):
-        """Y co-ordinate of the cartesian position."""
-        return self._y
-
-    @property
-    def z(self):
-        """Z co-ordinate of the cartesian position."""
-        return self._z
+CartCoord = namedtuple("CardCoord", ["x", "y", "z"])
```
PeterJCLaw: While the syntax (at least before Python 3.6) is ugly, I think it would be worth this being a `typing.NamedTuple`:

``` python
CartCoord = NamedTuple('CartCoord', [
    ('x', int),
    ('y', int),
    ('z', int),
])
```

robot/markers.py
```diff
@@ -83,20 +68,15 @@ class Marker:
...
         """ID of the marker seen."""
-        return self._id
+        return self._raw_data['id']
 
     @property
-    def size(self):
+    def size(self) -> Tuple[TMetres, TMetres]:
         """Marker size in metres."""
-        return tuple(self._size)
+        return self._raw_data['size']
```
PeterJCLaw: Does this not need conversion to a tuple? (I'm assuming that it's a list by default given that it's decoded from JSON)

robot/servo.py
```diff
@@ -70,7 +71,8 @@ def mode(self, mode: PinMode):
             raise ValueError("Mode should be a valid 'PinMode', got {!r}".format(mode))
         self._pin_mode_set(mode)
 
-    def read(self):
+
```
PeterJCLaw: Is this added newline intentional?

setup.cfg
```diff
@@ -36,3 +36,13 @@ multi_line_output = 3
...
+
+
+[mypy]
+warn_unused_ignores = True
+strict_optional = True
+
+disallow_any_generics = True
+check_untyped_defs = True
+
+ignore_missing_imports = True
```
PeterJCLaw: Is this needed? I'd greatly enjoy inverting this if possible.

robot/camera.py
RealOrangeOne: The functionality is the same, but I prefer to be explicit if we're extracting the exception as a variable anyway

PeterJCLaw: In Python 2.7 the functionality between the two was very different; I've not yet had enough of a play with Python 3 to know what the differences are now.

Even if thy are the same, I think I'll still be in favour of plain `raise` since that's the idiomatic way of saying "I'm re-raising the original exception, as if I'd not caught it" (which is what this code is trying to do).

PeterJCLaw: Hrm, so it turns out that in Python 3 we don't lose the underlying trace (like we did in Python 2), though there is a difference in the outcome from re-raising vs raising the original exception explicitly. Consider the following example:
``` python
import traceback

def func():
    raise Exception("from func")

def reraise():
    try:
        func()
    except Exception as e:
        raise

def raise_e():
    try:
        func()
    except Exception as e:
        raise e

for fn in (reraise, raise_e):
    print("\nTesting with {}".format(fn.__name__ ))
    try:
        fn()
    except:
        traceback.print_exc()
```
The output of this is:
```
> py foo.py

Testing with reraise
  File "foo.py", line 21, in <module>
    fn()
  File "foo.py", line 8, in reraise
    func()
  File "foo.py", line 4, in func
    raise Exception("from func")
Exception: from func

Testing with raise_e
  File "foo.py", line 21, in <module>
    fn()
  File "foo.py", line 16, in raise_e
    raise e
  File "foo.py", line 14, in raise_e
    func()
  File "foo.py", line 4, in func
    raise Exception("from func")
Exception: from func

> py --version
Python 3.6.4
```

setup.cfg
PeterJCLaw: Indeed, it's rather unfortunate that we need to ignore some imports there (notably I think it was cffi, numpy and PIL at least which don't have typeshed entries), however this project doesn't have any such dependencies, so I'd hope that we could enforce typing everywhere.

robot/camera.py
RealOrangeOne: Looks like we don't use the underlying exception (something I didn't know python2 did, which just sounds annoying). But seeing the stack trace there, it'd be nice not to show the additional `raise` statement. Will revert back to a plain `raise`.

RealOrangeOne:
@PeterJCLaw @prophile looking through all the changes here, there's a large amount of ambiguity the raw data that comes back from `robotd`. Do you know if there's any way we can type _explicitly_ what keys / values to expect in that data, and their types? Or this a limitation Python hasnt solved yet? 

Alternatively, we could replace the dictionary with a `NamedTuple`, which does support specifying keys and types as we need.

PeterJCLaw:
I'm not aware of a way to specify a dictionary with a given set of keys. The reality is that _somewhere_ has to deal with the fact that the JSON blogs are coming back from `robotd` are merely dictionaries and _don't_ have type information.

One approach might be to explicitly convert the JSON into `NamedTuple`s immediately and then wrap those in our helper classes as needed (is that what you had in mind?), though that feels pretty close to what we're already doing.

PeterJCLaw:
It turns out that `mypy` is experimenting with a way to have dictionaries with specified keys. It's called [TypedDict](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#typeddict).

robot/markers.py
```diff
@@ -1,30 +1,16 @@
 import math
+from collections import namedtuple
+from typing import List, NewType, Tuple
 
 from robot.game_specific import TOKEN, WALL
 
+TMetres = NewType('TMetres', float)
+TDegrees = NewType('TDegrees', float)
+TRadians = NewType('TRadians', float)
+TPixel = NewType('TPixel', float)
```
PeterJCLaw: nitpick: any reason that `Pixel` is singular, while the other unit-types are plural?

PeterJCLaw: Is it really intentional that `Pixel` is `float` rather than `int`? (Do we really have pixel values which are fractional?)

robot/markers.py
PeterJCLaw: Resolved in #59 by 1d04e79c2a52d814bdef88c0469a38931a58b079.

robot/camera.py
PeterJCLaw: Resolved in #59 by 960c531.

robot/game.py
PeterJCLaw: Resolved in #59 by 54becc903d1bf9259648f59c77f48fff5642e107.

setup.cfg
PeterJCLaw: This is resolved in #59 by 5a08e1a.

robot/robot.py
```diff
@@ -23,13 +23,13 @@ class Robot:
 
     ROBOTD_ADDRESS = "/var/robotd"
 
-    def __init__(self, robotd_path=ROBOTD_ADDRESS):
+    def __init__(self, robotd_path: str=ROBOTD_ADDRESS) -> None:
```
PeterJCLaw: Imo we should allow a `Path` here as well.

PeterJCLaw: Resolved in #59 by dfe6917.

robot/markers.py
PeterJCLaw: Hnng, it's worse than that -- there is no `'size'` key in this dictionary. Ughfg. I've opened #60 to remove this which we can merge later.

robot/markers.py
RealOrangeOne: hmm, youre right, it seems unlikely to need fractional pixels. Is this addressed in #59 ?

PeterJCLaw: no

PR #37, opened 2017-12-15
add zones for tokens

Adimote:


Adimote:
Merging this because it needs to be in pi-image yesterday.

PR #41, opened 2018-01-07
Add CI linting

PeterJCLaw:
This adds a _large_ number of docstrings as well as a number of other code tidyups.

There are not intended to be any behavioural changes here, though there are a couple of places where I've changed the spelling of some logic so that it fits better.

**Note**: this has a small number of failures left. I'm not sure how they should be fixed, partly as they (mostly) look like they're part of code which is now dead:
- #44 odd behaviour in [`Robot.close`](https://github.com/sourcebots/robot-api/commit/e5f824289369c3deeda35f949fe0782582962f71#r26705712) meaning I'm not sure what the method is meant to do
- ~~#42 dead `Servo_get_status` method which uses undefined variable~~
- #43 `zone_script.py`


robot/__init__.py
```diff
@@ -1,5 +1,7 @@
...
-from robot.game import GameMode
-from robot.game_specific import *
-from robot.servo import PinMode, PinValue
+"""Userspace API for a robot running ``robotd``."""
+
+from robot.robot import Robot  # noqa
+from robot.motor import BRAKE, COAST  # noqa
+from robot.game import GameMode  # noqa
+from robot.game_specific import *  # noqa
+from robot.servo import PinMode, PinValue  # noqa
```
prophile: Rather than `noqa`, we should make the export explicit with `__all__`.

PeterJCLaw:
@Adimote and/or @RealOrangeOne what's the status of `robot/zone_script.py`? It doesn't really fit with this project and appears untested (at best) and possibly broken.

RealOrangeOne:
@PeterJCLaw It's been tested locally by @Adimote, however not for a while (pre-smallpeice IIRC). I doubt it's broken, but it's highly likely in need of some improvement.

I would agree it should be in its own repo, with its own documentation, as it's never actually used in `robot-api`. I'll extract it out now into its own repo, so it can be better managed there.

robot/__init__.py
PeterJCLaw: The challenge is that that would then require duplicating all the game-specific definitions in this file. I'm not sure that's really something we want, though maybe is the best solution.

RealOrangeOne: If `__all__` is correctly evaluated, we could explicitly define everything else, and then unpack the variables from the `game_specific` module into `__all__`?

It's a hack, but it could work, and be _fairly_ nice?

PeterJCLaw: I don't think that's how `__all__` works -- you need to populate it with strings of symbol _names_.

PeterJCLaw: c9ef126 explicit is better than implicit. The overhead here is minimal (linting will keep us honest if we change things, though admittedly not if we add things), so I think this is best.

robot/robot.py
```diff
@@ -173,18 +201,20 @@ def zone(self):
...
-        competition or development mode respectively.
+        The ``GameMode`` the robot is in.
+
+        :return: one of ``GameMode.COMPETITION`` or ``GameMode.DEVELOPMENT``.
         """
         return self._game.mode
 
     def close(self):
+        """
+        Close all the baord connections this instance holds.
```
RealOrangeOne: Typo in docstring

PR #42, opened 2018-01-08
Remove dead Servo._get_status method

PeterJCLaw:
I'm not sure what this was used for, but it's unused.


kierdavis:
Looks like b4426cc453157ef3f9446cb0831c6e61bec6143e added another method `_get_servo_pos` that does more or less the same thing, and removed the only reference to `_get_status` without removing the `_get_status` method itself.

:+1: for removing.

PR #43, opened 2018-01-08
Remove the zone script

RealOrangeOne:
Extracted into https://github.com/sourcebots/zone-script

https://github.com/sourcebots/robot-api/pull/41#issuecomment-355864224


PeterJCLaw:
@RealOrangeOne thanks! Could we add a bit more documentation to the new repo? I'm also wondering whether the script itself might fit best into https://github.com/sourcebots/robotd; I think the challenge is that I'm not currently sure how it's meant to be used nor what dependencies it has.

PeterJCLaw:
The CI failure here is a result of there being a test for the `zone_script` which needs looking at.

tests/test_game.py
```diff
@@ -24,20 +23,6 @@ def test_default_state(self):
         self.assertEqual(self.robot.zone, 0)
         self.assertEqual(self.robot.mode, GameMode.DEVELOPMENT)
 
-    def test_game_state(self):
-        self.mock.new_gamestate()
-        zone = 2
-        self.thread = Thread(target=zone_script.poll, args=("/tmp/robotd/", zone, self.stop_event))
```
PeterJCLaw: I think this also makes some of the imports and instance attributes redundant (including the `tearDown` method)

PR #44, opened 2018-01-11
Explicit Robot and Board close handling

PeterJCLaw:
Tidy how the `Robot` class closes its `Board`s and how they are closed.

This fixes an issue where previously nothing actually happened in `Robot.close` (see my [comment](https://github.com/sourcebots/robot-api/commit/e5f824289369c3deeda35f949fe0782582962f71#r26705712) for details).

### DIFF
--- a/robot/robot.py
+++ b/robot/robot.py
@@ -180,6 +180,9 @@ class Robot:
         return self._game.mode
 
     def close(self):
+        """
+        Close all the baord connections this instance holds.
+        """
         # remove all the boards
         for board_type in self.all_known_boards:
             for board in board_type:
--- a/robot/robot.py
+++ b/robot/robot.py
@@ -183,8 +183,8 @@ class Robot:
         """
         Close all the baord connections this instance holds.
         """
-        for board_type in self.all_known_boards:
-            for board in board_type:
+        for board_group in self.all_known_boards:
+            for board in board_group:
                 board.close()
 
     def __del__(self):
--- a/robot/robot.py
+++ b/robot/robot.py
@@ -185,7 +185,7 @@ class Robot:
         """
         for board_type in self.all_known_boards:
             for board in board_type:
-                del board
+                board.close()
 
     def __del__(self):
         self.close()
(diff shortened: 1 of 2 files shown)


robot/robot.py
```diff
@@ -180,10 +180,16 @@ def mode(self):
...
+        """
+        Close all the baord connections this instance holds.
+        """
+        for board_group in self.all_known_boards:
+            for board in board_group:
+                board.close()
+
+            # Clear the group so that any further access doesn't accidentally
+            # reanimate the boards (which isn't supported).
+            del board_group[:]
```
RealOrangeOne: I'm not familiar with with `[:]` syntax, what's it doing?

PeterJCLaw: `[:]` without any values is an empty slice and is the extension of the partial slices `[:N]` and `[N:]` -- it returns an unbounded "slice" of the object (which is usually equal to the original object, though not the same instance).

For lists it can be used as a handy way to duplicate the list:
``` python
>>> foo = [0, 1, 2, 3]
>>> bar = foo[:]
>>> foo, bar
([0, 1, 2, 3], [0, 1, 2, 3])
>>> foo.append('foo-only')
>>> foo, bar
([0, 1, 2, 3, 'foo-only'], [0, 1, 2, 3])
>>> bar.append('bar-only')
>>> foo, bar
([0, 1, 2, 3, 'foo-only'], [0, 1, 2, 3, 'bar-only'])
```

Note that you can also assign into slices:
``` python
>>> foo = [0, 1, 2, 3]
>>> foo[1:2] = ['a', 'b']
>>> foo
[0, 'a', 'b', 2, 3]
```
This is handy as you can _clear_ a list (i.e: remove all the elements, but keep the same `list` instance) by doing `foo[:] = []`.

The idea of adding `del` to that, in order to make `del foo[:]` was new to me too, and does much the same thing -- it clears the list.

If you want to see how these work, the relevant magic methods are [`__getitem__`](https://docs.python.org/3/reference/datamodel.html#object.__getitem__), [`__setitem__`](https://docs.python.org/3/reference/datamodel.html#object.__setitem__) and [`__delitem__`](https://docs.python.org/3/reference/datamodel.html#object.__delitem__).

PR #45, opened 2018-01-11
Tidy vision results

PeterJCLaw:
This adds tests for the vision results and simplifies the code getting them.

### DIFF
--- a/robot/camera.py
+++ b/robot/camera.py
@@ -30,7 +30,6 @@ class Camera(Board):
             markers.append(Marker(token))
         # Sort by distance
         return ResultList(sorted(markers, key=lambda x: x.distance_metres))
-        return ResultList(markers)
 
     @property
     def serial(self):
--- a/robot/camera.py
+++ b/robot/camera.py
@@ -56,44 +56,20 @@ class Camera(Board):
                     raise
 
 
-class ResultList(MutableSequence):
+class ResultList(list):
     """
-    This class pretends to be a list, except it returns
-    a much more useful error description if the user indexes an empty array.
+    This class returns a more beginner-friendly error messages if the user
+    indexes into it, but it is empty.
 
     This is to mitigate a common beginners issue where an array is indexed
     without checking that the array has any items.
     """
 
-    def __delitem__(self, index):
-        del self.data[index]
-
-    def __init__(self, data):
-        self.data = data
-
-    def __getitem__(self, item):
+    def __getitem__(self, *args, **kwargs):
         try:
-            return self.data[item]
+            return super().__getitem__(*args, **kwargs)
         except IndexError as e:
-            if len(self.data) == 0:
+            if not self:
                 raise IndexError("Trying to index an empty list")
             else:
                 raise
-
-    def __setitem__(self, key, value):
-        self.data[key] = value
-
-    def insert(self, index, value):
-        self.data.insert(index, value)
-
-    def __len__(self):
-        return len(self.data)
-
-    def __repr__(self):
-        return self.data.__repr__()
-
-    def __eq__(self, other):
-        return self.data.__eq__(other)
-
-    def __iter__(self):
-        return self.data.__iter__()
--- a/robot/camera.py
+++ b/robot/camera.py
@@ -25,9 +25,10 @@ class Camera(Board):
         :param data: json string data to convert
         :return: #ResultList object (that imitates a list) of markers
         """
-        markers = [Marker(x) for x in data["markers"]]
-
-        return ResultList(sorted(markers, key=lambda x: x.distance_metres))
+        return ResultList(sorted(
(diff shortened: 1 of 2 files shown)


tests/test_camera.py
```diff
@@ -56,8 +59,13 @@ def test_can_see_something(self):
         camera = self.robot.cameras[0]
         tokens = camera.see()
 
-        # Check the correct markers are spotted
-        self.assertEqual({x.id for x in tokens}, {9})
+        self.assertEqual({x.id for x in tokens}, {9}, "Saw wrong markers")
+
+        self.assertEqual(
+            9,
+            tokens[0].id,
```
RealOrangeOne: Assuming we follow the standard of _not_ using [yoda conditions](https://en.wikipedia.org/wiki/Yoda_conditions), the first 2 args here should be reversed.

tests/test_camera.py
```diff
@@ -68,3 +76,77 @@ def test_unique_error(self):
...
+
+        with self.assertRaises(IndexError):
+            rl[5]
+
+    def test_many_items(self):
+        expected = ["spam", "ham"] * 3
+        rl = ResultList(expected)
+
+        self.assertEqual(6, len(rl))
+        self.assertEqual(expected, [x for x in rl])
```
RealOrangeOne: Surely as `ResultList` is a subclass of list, the list comprehension isnt needed here?

tests/test_camera.py
```diff
@@ -68,3 +76,77 @@ def test_unique_error(self):
...
+
+        self.assertEqual(expected[0], rl[0])
+        self.assertEqual(expected[1], rl[1])
+        self.assertEqual(expected[2], rl[2])
+        self.assertEqual(expected[3], rl[3])
+        self.assertEqual(expected[4], rl[4])
+        self.assertEqual(expected[5], rl[5])
```
RealOrangeOne: Are these explicit checks necessary? The previous `assertEqual` does the same thing under the hood, I believe

tests/test_camera.py
PeterJCLaw: Unfortunately unittest is rather unclear about this in its docs. While I agree with the yoda conditions comment, my experience is that the standard among unit testing frameworks is to have signatures like `assertEqual(expected, actual, message=None, *message_args)`.

This _is_ something which `unittest` conforms to, though you can only see it some of the time. Here's an example where it matters:
``` python
import unittest

class ExampleTest(unittest.TestCase):
    def test_exmaple(self):
        expected = {'expected': 'bar'}
        actual = {'actual': 'spam'}
        self.assertEqual(expected, actual)
```
which outputs:
```
F
======================================================================
FAIL: test_exmaple (goo.ExampleTest)
----------------------------------------------------------------------
  File "/home/sourcebots/robot-api/goo.py", line 7, in test_exmaple
    self.assertEqual(expected, actual)
AssertionError: {'expected': 'bar'} != {'actual': 'spam'}
- {'expected': 'bar'}
+ {'actual': 'spam'}

----------------------------------------------------------------------
Ran 1 test in 0.001s

FAILED (failures=1)
```
note the diff-like output of the expected vs actual.

If you put the expected and the actual the other way around, the message that you get would be rather confusing. I've unfortunately had a number of occasions (and quite a lot of time spent) where I've found that I'm investigating a failing test that I can't understand the reason for the failure only to find that the arguments to the assertion were back to front leading to incorrect output.

I therefore strongly prefer putting the expected first, even though a bit yoda it does read.

tests/test_camera.py
PeterJCLaw: This is to explicitly test iteration -- does 8d4628b3fa5cceb94d5b49f0903b89c3ef5db615 help clarify?

tests/test_camera.py
PeterJCLaw: This is to explicitly test indexing -- does 8d4628b3fa5cceb94d5b49f0903b89c3ef5db615 help clarify?

tests/test_camera.py
RealOrangeOne: I'm not sure how i've missed that after all this time... I know `assertIn` works like that, but as does the `in` operator, so i'm ok with that. 

Seeing as `==` works either way like this, I can live with it differing

robot/camera.py
```diff
@@ -56,44 +54,20 @@ def see(self):
                     raise
 
 
-class ResultList(MutableSequence):
+class ResultList(list):
```
RealOrangeOne: Arguably OOS of this PR, but it might be nice to make this immutable, by overriding `__setitem__` to raise an exception, or extending `Sequence`

PeterJCLaw: I'm not sure whether we need to do that (though I don't feel strongly). Agree it's out of scope for this PR.

PR #48, opened 2018-01-13
Build process cleanup

RealOrangeOne:
- Remove unnecessary comments, clearly left in from original template
- Pin versions of `robotd` and `sb-vision`
- Dont manually clone repos, `pip` will handle this for us


.circleci/config.yml
```diff
@@ -1,34 +1,20 @@
...
     working_directory: ~/robot
 
     steps:
       - checkout
 
-      # Download and cache dependencies
       - restore_cache:
           keys:
           - v1-dependencies-{{ checksum "requirements.txt" }}
           - v1-dependencies-{{ checksum "robotd/requirements.txt" }}
```
PeterJCLaw: I'm not sure this line can ever do anything, though even if it does I don't think we want that to happen. Remove it?

RealOrangeOne: Hmm, yea, agreed. Not sure what the use of that line really is

kierdavis:
@RealOrangeOne unfortunately this has conflicts now - if you can fix them I'd be happy to approve this.

.circleci/config.yml
```diff
@@ -1,34 +1,18 @@
...
-      # specify the version you desire here
-      # use `-browsers` prefix for selenium tests, e.g. `3.6.1-browsers`
-      - image: python:3-stretch
-
-      # Specify service dependencies here if necessary
-      # CircleCI maintains a library of pre-built images
-      # documented at https://circleci.com/docs/2.0/circleci-images/
-      # - image: circleci/postgres:9.4
-
+      - image: python:3.5-stretch
```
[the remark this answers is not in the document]
RealOrangeOne: Apparently, the downside is that it doesnt exist 😂 https://circleci.com/gh/sourcebots/robot-api/336. 

I think it's *always* better to pin as specific as possible, especially as theres lots of changes in `3.6`, which we don't support anyway. I'll find the true image name

.circleci/config.yml
```diff
@@ -1,58 +1,39 @@
...
           paths:
             - ./venv
           key: v1-dependencies-{{ checksum "requirements.txt" }}
 
-      # run tests!
       - run:
           name: run tests
           command: python3 -m unittest
-
+
```
Adimote: [nitpick]: trailing whitespace

.circleci/config.yml
```diff
@@ -1,58 +1,39 @@
-# Python CircleCI 2.0 configuration file
-#
-# Check https://circleci.com/docs/2.0/language-python/ for more details
-#
```
Adimote: Other than neatening, was there any other reason to remove this header? I think it's good to have a human readable section explaining this otherwise cryptic file.

.circleci/config.yml
RealOrangeOne: oh god, how the hell did that get there?! 😱

Adimote:
I think this is failing because its using a robotd that doesn't support custom commands...?

.circleci/config.yml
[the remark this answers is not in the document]
PeterJCLaw: Debian stretch is _not_ going to change its Python minor version. Pinning to a level more than `3-stretch` is somewhat pointless.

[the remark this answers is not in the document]
RealOrangeOne: Have changed back to `python:3-stretch`, as no other relevant tags existed

PR #53, opened 2018-01-27
Fix issues with making socket_path a Path

PeterJCLaw:
Following @kierdavis' [investigation](https://github.com/sourcebots/robot-api/pull/22#issuecomment-360309918) into the issues with changing the `socket_path` to being a `Path` this introduces types in a couple of places where `socket_path` is used in a non-trivial manner (i.e: not places where we're just passing it through to something else; 
6cb4ec7 and aadbaad), renames a variable for clarity (7bdc970) and then introduces a fix (959cf47).

I suggest reviewing by commit -- I've design them so that after reading the tidyups the fix should be obvious and self-explanatory.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ jobs:
     docker:
       # specify the version you desire here
       # use `-browsers` prefix for selenium tests, e.g. `3.6.1-browsers`
-      - image: python:3.4-jessie
+      - image: python:3-stretch
 
       # Specify service dependencies here if necessary
       # CircleCI maintains a library of pre-built images
--- a/robot/board.py
+++ b/robot/board.py
@@ -3,6 +3,7 @@ import socket
 import time
 from collections import Mapping
 from pathlib import Path
+from typing import Union
 
 
 class BoardList(Mapping):
@@ -30,7 +31,7 @@ class Board:
     SEND_TIMEOUT_SECS = 6
     RECV_BUFFER_BYTES = 2048
 
-    def __init__(self, socket_path):
+    def __init__(self, socket_path: Union[Path, str]):
         self.socket_path = Path(socket_path)
         self.socket = None
         self.data = b''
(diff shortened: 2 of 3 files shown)


kierdavis:
These changes make sense to me, but since I haven't gotten round to familiarising myself with the typing framework we're using I can't comment on whether these changes are acceptable.

Just to confirm what the fix (959cf47) is doing - presumably the odd behaviour I was seeing is being caused by a `Path` not being considered equal to a `str` representing the same filename? This would lead the [set difference operation](https://github.com/PeterJCLaw/robot-api/blob/7bdc97037aaf06d5c69680f7bd3d9a3feadcecab/robot/robot.py#L72) to return `new_paths` unmodified, since no item of `known_paths` (which are `Path`s) would compare equal to any element of `new_paths` (which are `str`s), hence breaking the board detection/removal logic.

:+1: for adding typing to the relevant parts of the code here - it makes the problem and its solution much more apparent.

I've copied your source branch to [sourcebots/PeterJCLaw--fix-serial-in-base](https://github.com/sourcebots/robot-api/commits/PeterJCLaw--fix-serial-in-base), which allows the [CI to automatically run on it](https://circleci.com/gh/sourcebots/robot-api/292?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link). It's complaining about "No module named 'typing'", something I can't offer comment on how to resolve unfortunately.

PeterJCLaw:
@kierdavis your analysis of the issue is correct.

Do we know why the CI doesn't run for PRs from different repos? (IMO they should be!)

The lack of a `typing` module is a result of the current default container being a 3.4 one; since the Pis are running Python 3.5 we should change that.

PeterJCLaw:
The remaining lint failures here appear unrelated to these changes.

RealOrangeOne:
> Do we know why the CI doesn't run for PRs from different repos? (IMO they should be!)

Turns out this is an opt-in feature, for some reason for CircleCI. I've now enabled this for this repo. As the tests do seem to be failing, i'm not sure what we should do about it?

PeterJCLaw:
@RealOrangeOne the failures on this branch are (I think) lint issues which are more relevant to the underlying `serial-in-base` branch; I think it's better to fix them there.

PR #54, opened 2018-01-27
Add overload declarations for `ResultList.__getitem__`

PeterJCLaw:
This clarifies that slicing a `ResultList` returns a `List[Marker]`.

While that may not quite be what we actually want, changing it is out of scope for now.

### DIFF
--- a/robot/camera.py
+++ b/robot/camera.py
@@ -2,7 +2,7 @@ from pathlib import Path
 from robot.board import Board
 from robot.markers import Marker
 import time
-from typing import List
+from typing import List, overload
 import socket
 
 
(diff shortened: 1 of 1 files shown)


robot/camera.py
```diff
@@ -17,7 +17,15 @@ class ResultList(List[Marker]):
...
 
+    @overload
     def __getitem__(self, index: int) -> Marker:
+        ...
+
+    @overload  # noqa: F811 (deliberately method replacement)
+    def __getitem__(self, index: slice) -> List[Marker]:
+        ...
+
+    def __getitem__(self, index):  # noqa: F811 (deliberately method replacement)
```
RealOrangeOne: Are we able to define this as `Union[int, slice]` explicitly, or does the type checker work that out? Currently passing anything other than `int` or `slice` will be allowed at the type level and return `Any`, I think?

PeterJCLaw: I don't know whether the checker would allow `Union[int, slice]` if we changed the return type to `Any`. I suspect that it would then still error that it doesn't match the parent type.

While it looks like duplication, I think that what's here is the right thing to have.

PR #55, opened 2018-01-28
Add support for generic arduino commands

PeterJCLaw:
This adds a new `direct_command` method which takes in a command as a list of JSON-encodable entities and passes that to robotd. It expects to immediately get back a response message (as well as the usual status blob) which contains either valid data from the arduino or a description of the failure.

This includes a novel testing approach which:
- creates a valid robotd Board instance
- mocks the underlying serial connection
- creates a robot-api Board instance
- mocks the socket connecting the robot-api Board such that it
  connects to the robotd Board instance

This avoids real `socket`s and `multiprocessing`, both of which incur substantial overhead during tests along with the introduction of timeouts (which is just nasty).

This relies on https://github.com/sourcebots/robotd/pull/35.

### DIFF
--- a/robot/servo.py
+++ b/robot/servo.py
@@ -1,5 +1,6 @@
 from enum import Enum
 from pathlib import Path
+from typing import Any, List, Tuple
 
 from robot.board import Board
 
@@ -83,6 +84,24 @@ class Gpio:
         return self._pin_read()
 
 
+class ArduinoError(Exception):
+    """Base class for exceptions fed back from the ``ServoBoard`` (arduino)."""
+
+    pass
+
+
+class CommandError(ArduinoError):
+    """The servo assembly experienced an error in processing a command."""
+
+    pass
+
+
+class InvalidResponse(ArduinoError):
+    """The servo assembly emitted an response which could not be processed."""
+
+    pass
+
+
 class ServoBoard(Board):
     """
     A servo board, providing access to ``Servo``s and ``Gpio`` pins.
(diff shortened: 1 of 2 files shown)


PeterJCLaw:
The test failure on the CI is due to https://github.com/sourcebots/robotd/pull/35 not being available in that test environment.

RealOrangeOne:
I'm liking this use of fake sockets, hopefully these changes can be backported into the remainder of the tests (separate PR) to reduce the oddness they bring

PeterJCLaw:
@RealOrangeOne indeed, I've some thoughts -> #57

robot/servo.py
```diff
@@ -119,6 +138,34 @@ def serial(self):
...
+        # consume the broadcast status
+        self.receive()
+
+        if status == 'ok':
+            return response['data']
+        else:
+            for cls in (CommandError, InvalidResponse):
+                if cls.__name__ == response['type']:
+                    raise cls(response['description'])
+
```
kierdavis: What happens if we receive a response with an invalid `type`?

PeterJCLaw: good point; I'll make that error harder.

PeterJCLaw: 023496d

kierdavis:
After testing on a kit last night I'm confident this works as intended. I would be grateful if the conflicts could be resolved before I approve.

PeterJCLaw:
@kierdavis conflicts resolved; please could you take another look?

PR #61, opened 2018-02-02
Clarify coordinates

PeterJCLaw:
This exposes the new "spherical" coordinates from https://github.com/sourcebots/sb-vision/pull/23.

### DIFF
--- a/robot/markers.py
+++ b/robot/markers.py
@@ -15,12 +15,43 @@ CartCoord = NamedTuple('CartCoord', (
     ('z', Metres),
 ))
 
+_SphericalCoord = NamedTuple('SphericalCoord', (
+    ('rot_x_radians', Radians),
+    ('rot_y_radians', Radians),
+    ('dist', Metres),
+))
 
-class PolarCoord:
+
+class SphericalCoord(_SphericalCoord):
     """
     Represents a point expressed in polar co-ordinates.
 
-    Strictly the space are spherical co-ordinates.
+    This co-ordinate space describes a point as a angles about the Cartesian x
+    and y axes and a distance from the observer.
+
+    Note: this co-ordinate space is differnt to the usual representation of a
+    spherical space which is less coupled to the Cartesian reference frame.
+    """
+
+    @property
+    def rot_x_degrees(self) -> Degrees:
+        """Rotation about the x-axis in degrees."""
+        return Degrees(math.degrees(self.rot_x_radians))
+
+    @property
+    def rot_y_degrees(self) -> Degrees:
+        """Rotation about the y-axis in degrees."""
+        return Degrees(math.degrees(self.rot_y_radians))
+
+
+class PolarCoord:
+    """
+    Deprecated: represents a point expressed in legacy "polar" co-ordinates.
+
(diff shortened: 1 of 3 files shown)


robot/markers.py
```diff
@@ -117,12 +149,18 @@ def is_token_marker(self) -> bool:
     @property
     def polar(self) -> PolarCoord:
         """
-        The position of the marker in the polar co-ordinates.
+        Deprecated: the position of the marker in legacy "polar" co-ordinates.
```
kierdavis: If `@deprecated` was part of the standard library I'd recommend using it here. However it seems it would require pulling in a [third party library](https://pypi.python.org/pypi/deprecation) and we're raising a warning manually anyway, so its up to you whether you want to do this or not - I'm happy either way.

robot/markers.py
```diff
@@ -133,3 +171,14 @@ def cartesian(self) -> CartCoord:
...
         return CartCoord(*self._raw_data['cartesian'])
+
+    @property
+    def spherical(self) -> SphericalCoord:
+        """
+        The position of the marker in Spherical co-ordinates.
+
+        This co-ordinate space describes a point as a angles about the Cartesian
+        x and y axes and a distance from the camera. Note: this co-ordinate
+        space is differnt to the usual representation of a spherical space.
```
kierdavis: differnt -> different

robot/markers.py
```diff
@@ -133,3 +171,14 @@ def cartesian(self) -> CartCoord:
         The camera's position is the origin of the co-ordinate space.
         """
         return CartCoord(*self._raw_data['cartesian'])
+
+    @property
+    def spherical(self) -> SphericalCoord:
+        """
+        The position of the marker in Spherical co-ordinates.
+
+        This co-ordinate space describes a point as a angles about the Cartesian
```
kierdavis: "as a angles"?

robot/markers.py
kierdavis: Perhaps its worth briefly clarifying what the "usual representation" is, and in what way our library differs from it?

robot/markers.py
```diff
@@ -15,12 +16,43 @@
...
+
+class SphericalCoord(_SphericalCoord):
     """
     Represents a point expressed in polar co-ordinates.
 
-    Strictly the space are spherical co-ordinates.
+    This co-ordinate space describes a point as a angles about the Cartesian x
+    and y axes and a distance from the observer.
+
+    Note: this co-ordinate space is differnt to the usual representation of a
```
Adimote: typo: different

robot/markers.py
PeterJCLaw: c944d4f

robot/markers.py
PeterJCLaw: c944d4f (fixes typo)

robot/markers.py
PeterJCLaw: c944d4f

robot/markers.py
PeterJCLaw: 3676b76 does this in the type docstring; I'd rather keep the property docstring short and related to the property rather than duplicating the type docstring.

PR #64, opened 2018-02-03
Run tests in separate temporary directories, and clean them up when done

kierdavis:


tests/mock_robotd.py
```diff
@@ -9,6 +12,20 @@
 from robotd.master import BoardRunner
 
 
+def create_root_dir():
+    return tempfile.mkdtemp(prefix="robot-api-test-")
+
+
+def remove_root_dir(dir):
```
PeterJCLaw: Can we please avoid shadowing a builtin here: `dir` -> `directory`.

tests/test_camera.py
```diff
@@ -25,12 +25,13 @@ class CameraTest(unittest.TestCase):
 
     # TODO add test for Serial number
     def setUp(self):
-        mock = MockRobotD(root_dir="/tmp/robotd")
+        self.root_dir = create_root_dir()
```
PeterJCLaw: There's a lot of repetition here, perhaps we could extract a mixin which handles all of this?

```python
class MockRobotDMixin:
    def create_robotd_mock(self):
        root_dir = tempfile.TemporaryDirectory(prefix="robot-api-test-")
        self.addCleanup(root_dir.cleanup)
        mock_robotd = MockRobotD(root_dir=root_dir.name)
        self.addCleanup(mock_robotd.stop)
        return mock_robotd
```
Usage looks like this:
```python
class MyTests(MockRobotDMixin, unittest.TestCase):
    def setUp(self):
        self.mock_robotd = self.create_robotd_mock()
        ...
        self.robot = Robot(robotd_path=self.mock_robotd.root_dir)
    ...
```

kierdavis: I like this - implemented in 8275d72.

tests/mock_robotd.py
kierdavis: No longer necessary as this function is removed by 8275d72.

PeterJCLaw:
@kier as I'm going to need to change the `setUp` methods a bit for #89, I'm going to merge this to master now (and then outwards to that branch),

kierdavis:
Thanks!

(this wasn't blocking on anything, I just hadn't gotten around to merging it yet)

PR #72, opened 2018-02-17
More types

PeterJCLaw:
This adds a bunch more types, though as a result we get more errors. I think some of the errors are fixable, though I'm not sure we can actually get to a point of having a clean type-check with some of the current structures. If things need changing, I'll probably do that in separate branches.

### DIFF
--- a/robot/robot.py
+++ b/robot/robot.py
@@ -1,12 +1,14 @@
 from pathlib import Path
-from typing import List, Set, Union  # noqa: F401
+from typing import Any, List, Set, Type, Union  # noqa: F401
 
-from robot.board import BoardList, TBoard
+from robot.board import Board, BoardList, TBoard  # noqa: F401
 from robot.camera import Camera
 from robot.game import GameMode, GameState, Zone
 from robot.motor import MotorBoard
 from robot.power import PowerBoard
 from robot.servo import ServoBoard
+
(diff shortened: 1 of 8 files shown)


robot/robot.py
```diff
@@ -139,7 +139,7 @@ def _games(self) -> BoardList[GameState]:
         return self._update_boards(self.known_gamestates, GameState, 'game')
 
     @staticmethod
-    def _single_index(name, list_of_boards: BoardList[TBoard]) -> TBoard:
+    def _single_index(name: Any, list_of_boards: BoardList[TBoard]) -> TBoard:
```
RealOrangeOne: Is there something we can constrain it down to besides `Any`?

PeterJCLaw: huh, that should, I think, be `str`? Not sure why I made that `Any`.

PeterJCLaw: Ah, I know why I made that `Any` -- because that's the technically correct type. The only thing we do with the name is format it intio a string, so we don't care what type it is.

RealOrangeOne: I think we should still constrain it to a better type that reflects what it should be, even if the implementation doesn't actually care what type it is

PeterJCLaw: c393d18

PR #73, opened 2018-02-17
Move to logging

PeterJCLaw:
This replaces all the print statements in the main code with suitable logging lines. This doesn't actually configure the logging yet, though we should probably do that.

### DIFF
--- a/tests/mock_robotd.py
+++ b/tests/mock_robotd.py
@@ -31,7 +31,7 @@ class MockBoardMixin:
         return self._status
 
     def command(self, cmd):
-        print("{} Command: {}".format(self._name, cmd))
+        print("{} Command: {}".format(self._name, cmd))  # noqa: T001
         self._status.update(cmd)
         self.message_queue.put(cmd)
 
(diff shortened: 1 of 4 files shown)


tests/mock_robotd.py
```diff
@@ -31,7 +31,7 @@ def status(self):
         return self._status
 
     def command(self, cmd):
-        print("{} Command: {}".format(self._name, cmd))
+        print("{} Command: {}".format(self._name, cmd))  # noqa: T001
```
RealOrangeOne: Is there a reason this print call was kept explicitly? Really it shouldve been changed to debug level if anything

PeterJCLaw: Only that I tend not to worry so much about prints in tests. There's not an obvious place to put a logger, nor configure logging.

If we switched to a test runner which captured logging output (and displayed it on failures), then I agree it's a bit more of an obvious change. (such test runners tend to also capture prints and treat them similarly, so it somewhat becomes moot).

PR #77, opened 2018-02-17
Actually configure logging

PeterJCLaw:
I jumped the gun a bit in #73 and didn't actually include any logging configuration before merging... oops.

### DIFF
--- a/robot/robot.py
+++ b/robot/robot.py
@@ -12,6 +12,10 @@ from robot.servo import ServoBoard
 LOGGER = logging.getLogger(__name__)
 
 
+def configure_logging() -> None:
+    logging.basicConfig(level=logging.INFO)
+
+
 class Robot:
     """
     Core class of the Robot API.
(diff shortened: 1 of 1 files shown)


RealOrangeOne:
Functionality which may be useful would be to enable teams to switch the robot into `DEBUG` logging mode, to get some additional information? Perhaps by passing `verbose_logging=True` in constructor? (Perhaps warrants more discussion?)

PeterJCLaw:
Maybe, though this comes back to the usual discussion of how much we wrap things in cotton wool for competitors vs just doing things well in a manner which an knowledgeable Python user would expect.

In the latter case, the competitors can very easily achieve `DEBUG` level logging as follows:
``` python
import logging
from robot import Robot

logging.basicConfig(level=logging.DEBUG)

r = Robot()
...
```

RealOrangeOne:
Hmm, ok, perhaps we should log the fact we're using `logging`, and that people can get more verbose logs by doing that, but that's outside this PR

PeterJCLaw:
@RealOrangeOne I've side-stepped the logging level issue in 6b4458f2ce10b4e0c80958038ece5a462abaddc9, (which I accidentally pushed to the wrong branch; I'd intended to add it to this PR; oh well).

PR #78, opened 2018-02-17
Rework robotd JSON connections

PeterJCLaw:
This is an attempt to both simplify the logic around the JSON sockets connecting robot-api to robotd and to prepare their usage for having type hints added.

The main unknown I'm facing is not really understanding what real-world scenarios each of the exceptions we currently catch relate to, and thus whether there is anything sensible that we can do in each case.

This branch begins to assume that if we reconnect a socket that we should recommence our instruction cycle with the backend, though doesn't apply this consistently. It's not clear to me how reliable this expectation is; in particular how it affect reconnections which occur in the middle of `_send_and_receive`.

### DIFF
--- a/robot/board.py
+++ b/robot/board.py
@@ -11,7 +11,7 @@ LOGGER = logging.getLogger(__name__)
 class Board:
     """Base class for connections to ``robotd`` board sockets."""
 
-    SEND_TIMEOUT_SECS = 6
+    CONNECTION_TIMEOUT_SECS = 6
 
     def __init__(self, socket_path: Union[Path, str]) -> None:
         self.socket_path = Path(socket_path)
@@ -40,7 +40,7 @@ class Board:
         :param socket_path: Path for the unix socket
         """
         self.socket = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
-        self.socket.settimeout(self.SEND_TIMEOUT_SECS)
+        self.socket.settimeout(self.CONNECTION_TIMEOUT_SECS)
 
         try:
             self.socket.connect(str(self.socket_path))
--- a/robot/board.py
+++ b/robot/board.py
@@ -1,9 +1,9 @@
-import json
 import logging
 import socket
-import time
 from pathlib import Path
 from typing import Mapping, TypeVar, Union
+
+from .connection import Connection, Message
 
 LOGGER = logging.getLogger(__name__)
 
@@ -15,10 +15,9 @@ class Board:
 
     def __init__(self, socket_path: Union[Path, str]) -> None:
         self.socket_path = Path(socket_path)
-        self.socket = None
-        self.data = b''
 
-        self._connect()
+        self.connection = Connection(self._get_socket())
+        self._greeting_response(self.connection.receive())
 
     @property
     def serial(self):
@@ -33,23 +32,28 @@ class Board:
         """
         pass
 
-    def _connect(self):
+    def _get_socket(self):
         """
         Connect or reconnect to a socket.
 
         :param socket_path: Path for the unix socket
         """
-        self.socket = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
-        self.socket.settimeout(self.CONNECTION_TIMEOUT_SECS)
+        sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+        sock.settimeout(self.CONNECTION_TIMEOUT_SECS)
 
         try:
-            self.socket.connect(str(self.socket_path))
+            sock.connect(str(self.socket_path))
         except ConnectionRefusedError as e:
             LOGGER.exception("Error connecting to: '%s'", self.socket_path)
             raise
 
-        greeting = self._receive()
-        self._greeting_response(greeting)
+        return sock
+
+    def _reconnect(self) -> None:
+        self.connection.close()
+
+        self.connection = Connection(self._get_socket())
+        self._greeting_response(self.connection.receive())
 
     def _get_lc_error(self) -> str:
         """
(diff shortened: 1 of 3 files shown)


robot/board.py
```diff
@@ -33,23 +32,28 @@ def _greeting_response(self, data):
         """
         pass
 
-    def _connect(self):
+    def _get_socket(self):
```
kierdavis: Could we keep the word "connect" in the name of this function, since that's the most important thing it's doing?

robot/board.py
```diff
@@ -15,10 +15,9 @@ class Board:
 
     def __init__(self, socket_path: Union[Path, str]) -> None:
         self.socket_path = Path(socket_path)
-        self.socket = None
-        self.data = b''
 
-        self._connect()
+        self.connection = Connection(self._get_socket())
```
kierdavis: This code is also duplicated in `_reconnect`. Perhaps it would be neater to initialise `connection` to `None` here and then immediately call `_reconnect`, and modify `_reconnect` to add a check for `connection is not None` around the call to `connection.close`?

robot/connection.py
```diff
@@ -0,0 +1,42 @@
...
+        self.socket.sendall(line)
+
+    def receive(self) -> Message:
+        """Receive a single message from the connection."""
+        while b'\n' not in self.data:
+            message = self.socket.recv(4096)
+            if message == b'':
+                raise BrokenPipeError()
+
+            self.data += message
```
kierdavis: Would it be worth using an [`io.BufferedReader`](https://docs.python.org/3/library/io.html#io.BufferedReader) here instead of implementing buffering ourselves? I haven't used this particular class before, but my intuition suggests we can just call `readline` on an instance of `BufferedReader` and it will deal with performing read operations on the socket only when needed.

However, your current approach has the advantage of being more explicit about its error handling. It's not clear without looking at its source how `BufferedReader` would handle an empty response from `socket.recv`.

robot/connection.py
```diff
@@ -0,0 +1,42 @@
+import json
+import socket
+from typing import Any, Dict
+
+Message = Dict[str, Any]
+
+
+class Connection:
```
kierdavis: It feels to me like "codec" would be a much better name for this class and module than "connection", especially since this class doesn't actually do any connecting (that's done in `Board._get_socket`).

robot/board.py
PeterJCLaw: I think I was going to have it _not_ connect at one point (just prime the socket), you're right that it's not great as it stands.
I do want it to be clear this is only part of the formation of a connection though. Maybe `_get_connect_socket`? (feels a bit verbose though).

robot/board.py
PeterJCLaw: This is deliberate -- I'm precisely moving away from having `None`s floating around the place. This is partly because they're annoying and partly because adding types exposes that.

robot/connection.py
PeterJCLaw: Interesting, I'll give that a look. I also found that we can turn a socket into a file-like object, though again I'm not sure how that handles errors.

As it stands, this class is pretty much copied from [robotd/master.py](https://github.com/sourcebots/robotd/blob/e0be841816d4c778e98498193f6fd7e3de35813d/robotd/master.py#L19-L53). If we do find that there's a good wrapper then we should consider using it there, too.

robot/connection.py
PeterJCLaw: Possibly, though it also does buffering which I don't think a codec would do. As it stands the name is merely the same as the class this is copied from in [robotd/master.py](https://github.com/sourcebots/robotd/blob/e0be841816d4c778e98498193f6fd7e3de35813d/robotd/master.py#L19-L53). I think it would be good to keep the two classes having the same name, but I don't feel strongly about what the name actually is.

PeterJCLaw:
@kierdavis thanks! This is definitely in need of some actual testing, as well as a better understanding of some of the errors it's now letting bubble straight away.

I'll update the description to clarify my goals as I think there's an even better way of managing the connections, that we've just not yet found.

robot/board.py
kierdavis: I think something like `_get_connected_socket` would be fine. This is very much a "write-once, read-often" piece of code, so I'd value a descriptive-but-accurate name over a brief one.

robot/board.py
kierdavis: Makes sense. In that case, it could be factored out into a new method that returns the new `Connection`?

kierdavis: e.g.

```python
def _connect(self):
  new_connection = Connection(self._get_socket())
  self._greeting_response(new_connection.receive())
  return new_connection

def __init__(self):
  # ...
  self.connection = self._connect()

def _reconnect(self):
  self.connection.close()
  self.connection = self._connect()
```

Assuming `self.connection` is not modified outside of the above functions, it will never be `None`.

In theory `_greeting_response` would need to be modified Assuming the above code is the olAssuming the above code is the olto use `new_connection` rather than `self.connection`, though since it's now only used once it could just be inlined into `_connect`. However I've just seen that our implementation of `_greeting_response` is simply a no-op, so I think it'd be clearer to just delete it (since clearly robotd does not require nor accept a response to the greeting packet it sends).

robot/connection.py
kierdavis: Tangentially related (and definitely out of scope), but a long-term improvement could be to factor out this logic from both robotd and robot-api into a library used by both (e.g. "robot-protocol"). This would facilitate writing other applications that interact with robotd, such as a manual control interface or the fabled Power Board Orchestra.

robot/connection.py
kierdavis: I agree that it would be good for the two classes to have the same name; as a result, I think we should rename both of them. Let's leave that to another PR, though.

robot/board.py
PeterJCLaw: `_greeting_response` is, I _think_, provided so that `Board` implementations can handle the initial status broadcast from `robotd`. You're right that it's unused at the moment.

I'll have a look at the rest later.

robot/connection.py
PeterJCLaw: Yup, a common library was another thing I'd wondered about here. I think we'd want to pin down some even higher level things before/while doing that (such as a generic way to pass errors back & forth) too.

I'd hope that something like a manual control interface would sit above the robot-api though -- is there a reason you think it needs to be a sibling?

kierdavis: >I'd hope that something like a manual control interface would sit above the robot-api though

My bad - that would make much more sense that using the hypothetical `robot-protocol` package directly.

Come to think of it I'm not sure what applications, if any, *would* use `robot-protocol` rather than `robot-api` in this sense.

PR #79, opened 2018-02-17
Simplify board lists

PeterJCLaw:
This changes `_update_boards` so that it handles all the logic needed by the various boards properties and resolves the TODO around that. It also moves more of the logic around the construction of `BoardList` into that class, making it simpler to create and to add types for.

### DIFF
--- a/robot/robot.py
+++ b/robot/robot.py
@@ -84,7 +84,7 @@ class Robot:
                     exc_info=True,
                 )
 
-        return sorted(boards, key=lambda b: b.serial)
+        return boards
 
     @staticmethod
     def _dictify_boards(boards: List[TBoard]) -> BoardList[TBoard]:
--- a/robot/robot.py
+++ b/robot/robot.py
@@ -64,27 +64,27 @@ class Robot:
         known_boards: List[TBoard],
         board_type: Type[TBoard],
         directory_name: _PathLike,
-    ) -> List[TBoard]:
+    ) -> BoardList[TBoard]:
         """
         Update the number of boards against the known boards.
 
-        :param known_boards: The list of all currently known boards.
+        :param known_boards: The list of all currently known boards; this list
+                             will be updated with any newly found boards.
         :param board_type: The type of board to create.
         :param directory_name: The relative directory to look in for new boards.
-        :return: A list of all the known boards (both previously known and newly
-                 found).
+        :return: A ``BoardList[TBoard]`` of all the known boards (both
+                 previously known and newly found).
         """
         known_paths = {x.socket_path for x in known_boards}  # type: Set[Path]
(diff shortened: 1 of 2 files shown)


robot/robot.py
```diff
@@ -57,89 +59,79 @@ def _assert_has_power_board(self):
         if not power_boards:
             raise RuntimeError('Cannot find Power Board!')
 
-    def _update_boards(self, known_boards, board_type, directory_name):
+    def _update_boards(
```
RealOrangeOne: Perhaps taking typing too far, but it'd be nice to use a type variable for this over any `TBoard`. Therefore we can specify that the known boards, the board type, and the resulting board list are all of the same type

PeterJCLaw: I'm not sure I follow -- my understanding is that that's what this does?

Modifying a usage to show the breakage:
``` python
    @property
    def motor_boards(self) -> BoardList[MotorBoard]:
        """
        :return: A ``BoardList`` of available ``MotorBoard``s.
        """
        return self._update_boards(self.known_motor_boards, PowerBoard, 'motor')
        # [mypy] error:Argument 2 to "_update_boards" of "Robot" has incompatible type "Type[PowerBoard]"; expected "Type[MotorBoard]"
```

PeterJCLaw: Thinking about this a bit more, are you suggesting that `_update_boards` end up looking a bit more like this:
``` python
    def _update_boards2(self, board_type: Type[TBoard]) -> BoardList[TBoard]:
        return self._update_boards(self.known_boards[board_type], board_type, board_type.directory)
```
If so, I actually wondered about something like that.

The issue is that there doens't seem to be a way to specify the type of the `known_boards` `dict` which it would need. In theory it would be something like `Dict[Type[TBoard], List[TBoard]]`, however that doesn't work like you'd hope it would -- the TBoard needs to be a single value for the whole type, rather than being dynamic over the key/value pairs in the `dict`.

I _think_ you could do something a bit like that with either a function in the place of the `known_boards` mapping or maybe a custom type, though either solution ends up adding more code than we'd really want for the resulting gain. If we were doing a _lot_ of operations like that then it might be worth it, but I don't think we're really on that scale here.

I do think there's room to have the directories specified on the types (which I've only just spotted), though I think that's outside the scope of this PR.

PeterJCLaw: `TBoard` ensures that all the arguments to `_update_board` are of the same kind of `Board`. Its return type then couples that to being the right type for the property.

I'm still not sure I understand what your concern/suggestion is here -- could you provide an example (probably code/psuedocode) of either a broken scenario you're trying to prevent, or of what usage of the better API might look like?

RealOrangeOne: By that statement, this is probably fine 👍

PR #82, opened 2018-02-24
Improve the docstrings for polar co-ordinates

Adimote:
Currently they're `#TODOs`

### DIFF
--- a/robot/markers.py
+++ b/robot/markers.py
@@ -65,37 +65,44 @@ class PolarCoord:
     @property
     def rot_x_rad(self) -> Radians:
         """
-        Rotation of marker relative to camera in the #TODO axis.
+        Rotation of marker around the x axis relative to camera in radians.
+
+        A positive angle represents a marker above the centre of the camera's vision.
         """
         return self._rot_x_rad
 
     @property
     def rot_y_rad(self) -> Radians:
         """
-        Rotation of marker relative to camera in the #TODO axis.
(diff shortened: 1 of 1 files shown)


PeterJCLaw:
The polar coordinates are deprecated and undocumented. I'm not sure I see the point in this?

PeterJCLaw:
See https://github.com/sourcebots/docs/pull/67#discussion_r165389391 for the previous investigation into the current `polar` coordinates; I've summarised the interesting part below since I suspect that comment will be collapsed if you click through.

`sb-vision`'s `polar_x` and `polar_y` are calculated by [`cartesian_to_polar`](https://github.com/sourcebots/sb-vision/blob/57675c144b89eedb7ef451483c5a69412bc44d4a/sb_vision/tokens.py#L161) and translate to `rot_x...` and `rot_y...` respectively on the `robot-api` side. These values are not really in the right pace to be used (they're the wrong side of the triangle) and they also express a "rotation along the {x,y} axis" rather than "rotation about the {x,y} axis".

vision-reference-current-polar-implementation

kierdavis:
Oops, I assumed this PR was regarding spherical coordinates rather than polar when I looked over it earlier.

I concur with @PeterJCLaw here - now that spherical is shipped and polar is officially deprecated, why are we still caring about this?

Adimote:
Good point, I're replaced them all with Deprecated

PR #83, opened 2018-02-25
Add Orientation

Adimote:
Add Orienations to markers, to expose the feature added in https://github.com/sourcebots/sb-vision/pull/34


robot/markers.py
```diff
@@ -46,6 +46,45 @@ def rot_y_degrees(self) -> Degrees:
...
+
+
+class Orientation(_Orientation):
+    """
+    Represents the orientation in 3d space as rotations around x, y, and z axes.
+
+    Rotations around the different axes can be thought of as follows:
+      - X rotation represents pitch. Positive rotation
+      around the X axis represents a rotation with the nearest point going downwards,
+      and furthest going upwards. (i.e. a bowing motion is a positive x rotation)
```
PeterJCLaw: It took me a couple of attempts to understand the second sentence here ("Positive rotation..."). I like that that approach applies quite easily to each of the descriptions, though it's a bit complex on first reading.

I wonder whether moving the example first would help people grok the idea sooner so that the technically-correct explanation then has an idea to hold onto?

Alternatively, is there a slightly different phrasing which is easier to grok first time?

(I'm assuming we'll use a similar phrasing in the main docs too, so I think this is worth getting right here)

robot/markers.py
```diff
@@ -46,6 +46,45 @@ def rot_y_degrees(self) -> Degrees:
...
+      - Y rotation represents yaw. Positive rotation around the Y axis represents
+      a rotation with the nearest point going to the left, and the furthest going to the
+      right.
+      - Z rotation represents roll. Positive rotation around the Z axis represents
+      a rotation with the highest point going to the right, and the bottom going to the
+      left.
```
PeterJCLaw: Z rotation is aligned with a clockwise rotation about the axis (when viewed from the positive end of the axis), yet the other rotations are counter-clockwise. Is the difference intentional?

I'm not sure whether this matters, just noting it as my assumption was that they'd be consistent.

robot/markers.py
```diff
@@ -46,6 +46,45 @@ def rot_y_degrees(self) -> Degrees:
...
+_Orientation = NamedTuple('Orientation', (
+    ('rot_x_radians', Radians),
+    ('rot_y_radians', Radians),
+    ('rot_z_radians', Radians),
+))
+
+
+class Orientation(_Orientation):
+    """
+    Represents the orientation in 3d space as rotations around x, y, and z axes.
```
PeterJCLaw: What is the order of application of the angles, and/or do they always apply to the cartesian axes of the camera?

A camera-relative interpretation of `(45°, 90°, 0)` is as follows: consider a marker which is leaning forwards by 45° but placed on a horizontal rotating platform (like a "lazy susan"); then rotate the platform by 90°.

A marker-relative interpretation is as follows: consider a marker perpendicular to a rotating platform which is angled 45° towards the observer (like a lazy susan on the surface of a table which has been half-pushed over); then rotate the platform by 90°.

In both scenarios the marker is now facing off to the side, however in the former its baseline is still horizontal while in the latter its baseline is at a 45° to the horizontal.

I'm not sure which of the two systems is easier to think about, nor which we've got data for. We should find out and document this however.

Note that if we apply the angles in the reverse order (`z` then `y` then `x`) then the two scenarios above change roles.

robot/markers.py
```diff
@@ -168,6 +207,18 @@ def spherical(self) -> SphericalCoord:
...
 
+    @property
+    def orientation(self) -> Orientation:
+        """
+        The rotation of the marker in relative to the camera.
+
+        Describes a rotation as angles about the x, y, and z axes, which
+        corresponds pitch, yaw, and roll of the marker respectively. The
+        angles are measured as an offset from the marker directly facing
+        the camera.
```
PeterJCLaw: I think the last sentence here should have "a marker" rather than "the marker".

robot/markers.py
Adimote: The Z axis actually comes towards you, so its still a counter-clockwise rotation around the axis.

PeterJCLaw: > The Z axis actually comes towards you,

You're right that I was wrong, though I'm not sure I agree with this statement -- positive Z is away from the camera.

> so its still a counter-clockwise rotation around the axis.

Indeed.

robot/markers.py
PeterJCLaw: @Adimote I see this is partially resolved by 6018425c576e63f34063727ab17e0d0b9c786e31, though we still don't describe whether the rotations are relative to the marker or the camera.

I _think_ I hope that they're relative to the fixed frame of the camera (I'm expecting that that would make the angles for a pair of markers more easily comparable), though I've not explore enough examples to know which is better/easier to think about.

robot/markers.py
PeterJCLaw: 4e59cbb

robot/markers.py
PeterJCLaw: ca54c3e

robot/markers.py
PeterJCLaw: Further to my being wrong here -- I think that it's `Y` which is actually clockwise; might this be a result of our y-coordinate being defined with a different positive direction to that of OpenCV?

robot/markers.py
Adimote: Took me a while to figure this out, but I'm pretty sure the rotations are actually applied relative to the marker. I've added this detail in 7535af1202c33c19a20484e90e379cde0c357f4d This means an upside-down marker would report inverse `rot_y`. We might want to investigate how to make this easier for the students to use in a later update.

Adimote:
Marked this as WIP because it's not ready for shipping due to the confusing output values

PR #85, opened 2018-03-02
Add version number log message

Adimote:
Print the version number when we start.

It grabs the version number from setup.py

### DIFF
--- a/robot/robot.py
+++ b/robot/robot.py
@@ -1,8 +1,8 @@
 import logging
 from pathlib import Path
 from typing import List, Set, Type, Union  # noqa: F401
-import pkg_resources  # part of setuptools
 
+import pkg_resources  # part of setuptools
 from robot.board import BoardList, TBoard
 from robot.camera import Camera
 from robot.game import GameMode, GameState, Zone
@@ -15,6 +15,7 @@ _PathLike = Union[str, Path]
 LOGGER = logging.getLogger(__name__)
 
 VERSION = pkg_resources.require("robot")[0].version
+
 
 def configure_logging() -> None:
     """
--- a/robot/robot.py
+++ b/robot/robot.py
@@ -53,7 +53,7 @@ class Robot:
 
         configure_logging()
 
-        LOGGER.info("Robot (v{}) Intialising...".format(__VERSION__))
+        LOGGER.info("Robot (v{}) Initialising...".format(__VERSION__))
         self._assert_has_power_board()
         self.power_board.power_on()
 
--- a/robot/robot.py
+++ b/robot/robot.py
@@ -1,6 +1,7 @@
 import logging
 from pathlib import Path
 from typing import List, Set, Type, Union  # noqa: F401
+import pkg_resources  # part of setuptools
 
 from robot.board import BoardList, TBoard
 from robot.camera import Camera
@@ -13,6 +14,7 @@ _PathLike = Union[str, Path]
 
 LOGGER = logging.getLogger(__name__)
 
+VERSION = pkg_resources.require("robot")[0].version
 
 def configure_logging() -> None:
     """
@@ -53,6 +55,7 @@ class Robot:
         configure_logging()
 
         LOGGER.info("Initializing Hardware...")
+        LOGGER.info("Robot Intiailised (v{})".format(VERSION))
         self._assert_has_power_board()
         self.power_board.power_on()
 
--- a/robot/robot.py
+++ b/robot/robot.py
@@ -53,8 +53,7 @@ class Robot:
 
         configure_logging()
 
-        LOGGER.info("Initializing Hardware...")
-        LOGGER.info("Robot Intiailised (v{})".format(__VERSION__))
+        LOGGER.info("Robot (v{}) Intialising...".format(__VERSION__))
         self._assert_has_power_board()
         self.power_board.power_on()
 
--- a/robot/robot.py
+++ b/robot/robot.py
@@ -2,18 +2,17 @@ import logging
 from pathlib import Path
 from typing import List, Set, Type, Union  # noqa: F401
 
+from robot import __VERSION__
 from robot.board import BoardList, TBoard
 from robot.camera import Camera
 from robot.game import GameMode, GameState, Zone
 from robot.motor import MotorBoard
 from robot.power import PowerBoard
 from robot.servo import ServoBoard
-from robot import __VERSION__
 
 _PathLike = Union[str, Path]
 
 LOGGER = logging.getLogger(__name__)
-
 
 
 def configure_logging() -> None:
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -2,7 +2,7 @@
 
 setup(
     name='robot',
-    version='1.0.0',
```
RealOrangeOne: I believe we agreed the version number would also contain the year, ie `2018.2.0` (as this is second release this year)

robot/robot.py
```diff
@@ -2,6 +2,7 @@
 from pathlib import Path
 from typing import List, Set, Type, Union  # noqa: F401
 
+import pkg_resources  # part of setuptools
```
RealOrangeOne: Is this comment especially necessary?

PeterJCLaw:
An alternative approach I've seen which works well (and avoids the need to import `pkg_resources`) is to define `__VERSION__` in the main package `__init__.py` and import _that_ into the `setup.py`. That does mean you need to have the dependencies installed to build the package, but since we already expect to build the tests as part of building the (debian) package, I think that's reasonable.

Other questions:
- what's the versioning strategy going to be? -- will we:
  - increment -> develop -> release, or
  - develop -> increment -> release
- how will the version numbers be defined? (there seems to be some disagreement about the format)
- how will we ensure that we always increment the version number before a release?
- can we feed the version number through to the debian package? (this would solve an outstanding issue where updates don't install)

robot/robot.py
```diff
@@ -53,6 +55,7 @@ def __init__(
         configure_logging()
 
         LOGGER.info("Initializing Hardware...")
+        LOGGER.info("Robot Intiailised (v{})".format(__VERSION__))
```
RealOrangeOne: I suggest either we move this down the initialization process, so the robot has actually initialized when we say it is. Or we don't say it's initialised, and move printing the version right to the top of the initialization. I'm fond of the latter

robot/robot.py
```diff
@@ -8,12 +8,14 @@
...
 from robot.power import PowerBoard
 from robot.servo import ServoBoard
+from robot import __VERSION__
 
 _PathLike = Union[str, Path]
 
 LOGGER = logging.getLogger(__name__)
 
 
+
```
RealOrangeOne: This additional new line shouldnt be there. There should be 2 blank lines between top-level constructs

Adimote: oops, fixed in c5123303adc7cd9f4d947820a623450ca01cf0e3

robot/robot.py
Adimote: fixed in 4b51ef0cb9879fa86d101334bfd7478b7cc02459 also the british english spelling is 'Initialise' :wink:

setup.py
Adimote: fixed in 8970025 (which I now realise has a typo in the commit message)

PeterJCLaw:
Another question, which I think has been asked before though I can't recall the answer:
- how will we cope with mixed versions of the software being used? This outputs the version of `robot-api` and doesn't say anything about the other things we install on the robot

Adimote:
@PeterJCLaw  re: your questions:
>  what's the versioning strategy going to be?
 I think we should go for develop -> increment -> release
> how will the version numbers be defined? (there seems to be some disagreement about the format)
Everyone else is right that we agreed to go for `2018.2.0`
> How will we ensure that we always increment the version number before a release?
We should make a release checklist
> can we feed the version number through to the debian package? (this would solve an outstanding issue where updates don't install)

> how will we cope with mixed versions of the software being used? This outputs the version of robot-api and doesn't say anything about the other things we install on the robot
I think to simplify this we should just not support mixed versions of the software being used. they would either have an update fail or consciously update only one system to do so.

robot/robot.py
```diff
@@ -52,7 +53,7 @@ def __init__(
 
         configure_logging()
 
-        LOGGER.info("Initializing Hardware...")
+        LOGGER.info("Robot (v{}) Intialising...".format(__VERSION__))
```
trickeydan: Typo here

Adimote: [not a typo](https://en.oxforddictionaries.com/definition/initialize)

trickeydan: [I'm afraid that it is](https://en.oxforddictionaries.com/search?utf8=%E2%9C%93&filter=dictionary&query=Intialising)

Adimote: but https://en.oxforddictionaries.com/search?utf8=✓&filter=dictionary&query=Intializing

trickeydan: Perhaps you mean "In**i**tializing" or "In**i**talising"?

Adimote: https://wikidiff.com/initializing/initialising settles this dispute completely.

PeterJCLaw:
@Adimote thanks! Here are some further thoughts:

>> How will we ensure that we always increment the version number before a release?
> 
> We should make a release checklist

Where is said checklist?

Unfortunately humans are fallible so we either need to make this _now_, before our first update release or, better, automate it. A strategy I've seen work well is that releases are deployed by pushing a tag. I'm not sure how well that would work with also actually making changes in the repo, though maybe a script which patches the repo, commits the result and pushes it up to a branch?

>> how will we cope with mixed versions of the software being used? This outputs the version of robot-api and doesn't say anything about the other things we install on the robot
> 
> I think to simplify this we should just not support mixed versions of the software being used. they would either have an update fail or consciously update only one system to do so.

Claiming not to support that scenario is one thing, actually being able to follow through is another. If a competitor installs mismatched versions how will we know that that's happened so that we can ignore their claims that things are broken? It's no use claiming that this is unsupported unless we can actually detect that to filter it out.

An update failing is (hopefully) unlikely, though again the lack of being able to detect that that has happened will greatly complicate things.

robot/robot.py
PeterJCLaw: @Adimote @trickeydan some general feedback here -- spotting typos in ones own writing is much harder than would be expected, even when you're told that there's a typo present.

I've found it much easier to fix this sort of error when I'm told _exactly_ what the issue is and/or had the suggested change outlined.

Here, this might have looked like:
> Typo: s/Intialising/Initialising/ (add missing second "i")

Similarly when pushing back on a review comment, including the detail of _why_ you feel the comment is wrong or doesn't need to be addressed can really help.

Here this might have looked like:
> I much prefer spelling "intialising" with an "s"; [the OED notes](https://en.oxforddictionaries.com/definition/initialize) that it's a valid spelling too.

Bonus: you might enjoy adding a spellchecker to your code editors. In VSCode I use https://marketplace.visualstudio.com/items?itemName=swyphcosmo.spellchecker; I'm sure similar plugins exist for other editors.

PR #86, opened 2018-03-06
add __repr__ for Board and Markers

Adimote:
Currently Board and Markers don't show pretty prints if you `print(markers)`. They do have `__str__()` functions, but `print()` calls `__repr__()`, so implement `__repr__()`

### DIFF
--- a/robot/board.py
+++ b/robot/board.py
@@ -155,6 +155,9 @@ class Board:
         """
         self.socket.detach()
 
+    def __repr__(self):
+        return "<{}>".format(self.__str__())
+
     def __str__(self):
         return "{} - {}".format(self.__name__, self.serial)
 


robot/markers.py
```diff
@@ -168,9 +168,12 @@ def spherical(self) -> SphericalCoord:
         """
         return SphericalCoord(*self._raw_data['spherical'])
 
+    def __repr__(self):
+        return "<robot.markers.{} >".format(self.__str__())
```
PeterJCLaw: You shouldn't need to call `__str__` explicitly here.

robot/board.py
```diff
@@ -155,6 +155,9 @@ def close(self):
         """
         self.socket.detach()
 
+    def __repr__(self):
+        return "<{}>".format(self.__str__())
```
PeterJCLaw: You shouldn't need to call `__str__` explicitly here.

robot/markers.py
Adimote: fixed in f6b302caa0584e038006c46df4e43ae474f7910e

robot/board.py
Adimote: fixed in f6b302caa0584e038006c46df4e43ae474f7910e

robot/markers.py
```diff
@@ -168,9 +168,12 @@ def spherical(self) -> SphericalCoord:
...
+    def __repr__(self):
+        return "<robot.markers.{} >".format(str(self))
+
     def __str__(self):
         bearing = self.spherical.rot_y_degrees
-        return "<Marker {}: {:.0f}° {}, {:.2f}m away>".format(
+        return "Marker {}: {:.0f}° {}, {:.2f}m away".format(
             self.id,
             abs(bearing),
             "right" if bearing > 0 else "left",
```
trickeydan: Do we need to specify "left" and "right" here if this is a bearing?

Also, I think clockwise and anti-clockwise might be more appropriate here as they are less ambiguous.

robot/markers.py
```diff
@@ -168,9 +168,12 @@ def spherical(self) -> SphericalCoord:
         """
         return SphericalCoord(*self._raw_data['spherical'])
 
+    def __repr__(self):
+        return "<robot.markers.{} >".format(str(self))
```
RealOrangeOne: I think replacing `robot.markers.` with nothing will lead to a nicer output. `<Marker 1: ... m away>` looks quite nice

robot/markers.py
Adimote: I think the issue here is that 'bearing' is the wrong variable name. I don't like mentioning clockwise or counter clockwise, as it's given without context here, and they might think it's the orientation of the marker

kierdavis:
IMO `__repr__` should aim to produce a string that could be passed back to `exec` to reproduce the object - if you want a pretty string that's what `__str__` is for. I'm open to objections though.

Anyone know if there's a PEP for the recommended behaviour of `__repr__`?

robot/markers.py
Adimote: either way, this isn't changed in this PR and thus should be fixed separately

kierdavis: I think `bearing` is probably fine here, if not 100% precise. The variable holds the bearing of the marker relative to the bearing of the camera. I feel using "left/right" to indicate "to the left/right of the camera" is acceptable here.

Adimote:
@kierdavis I tend to agree, however in this case it would be a really long string. I believe the agreed method for this is to try for something which can be passed to `exec()`, but if it's not possible it's fine to do your own thing

trickeydan:
From Python 3 docs:  For many types, this function(__repr__) makes an attempt to return a string that would yield an object with the same value when passed to eval(), otherwise the representation is **a string enclosed in angle brackets that contains the name of the type of the object together with additional information often including the name and address of the object.**

Adimote:
One thing I do think should always happen though is the output of `__repr__` should be unique for the reading. I don't think this means we need to give a hash of the object, but it should produce accurate enough numbers for slightly different readings to output a different `__repr__`

robot/board.py
```diff
@@ -155,8 +155,11 @@ def close(self):
         """
         self.socket.detach()
 
+    def __repr__(self):
+        return "<{}>".format(str(self))
```
PeterJCLaw: Please drop redundant `str` call here.
``` python
>>> class Foo:
...   def __repr__(self):
...     return 'repr'
...   def __str__(self):
...     return 'str'
... 
>>> print('{}'.format(Foo()))
str
```

robot/markers.py
```diff
@@ -168,9 +168,12 @@ def spherical(self) -> SphericalCoord:
         """
         return SphericalCoord(*self._raw_data['spherical'])
 
+    def __repr__(self):
+        return "<{}>".format(str(self))
```
PeterJCLaw: Please drop redundant `str` call here (as above).

Adimote: half-fixed (& broken something unrelated) in 7f9cfee 
Then undid the stupid breakage in 515e0a3 (not sure what I was on when I thought I fixed that.)
Then the other half is fixed in d4e364e

robot/board.py
```diff
@@ -155,8 +155,11 @@ def close(self):
         """
         self.socket.detach()
 
+    def __repr__(self):
+        return "<{}>".format(self)
```
PeterJCLaw: I suggest that this be:
``` python
    return '{}({!r}'.format(self.__class__.__name__, self.serial)
```
That would be truer `repr` behaviour, though would admittedly rely on classes which take other construction parameters overriding the default.

If we do that, we can then drop the `__str__` entirely.

trickeydan: I disagree, this would not be truer `repr` behaviour. Please see my above comment and link to the Python 3 documentation.

PeterJCLaw: Damn. Apparently I meant `self.socket_path`, not `self.serial`.

PeterJCLaw: In which case we might want to keep the `__str__`.

PeterJCLaw:
@Adimote was there a particular use-case or specific instance where you had seen a non-useful stringified board instance which this was trying to fix?

If not, I suggest we shelve this until after the competition.

PR #87, opened 2018-03-07
Bump version to 2018.2.0

Adimote:


Adimote:
@RealOrangeOne rebasing to master fixed the __VERSION__ bump issue

Adimote:
(as in, rebasing to master has added the "2018.2.0" to the `__VERSION__` value)

debian/changelog
```diff
@@ -3,3 +3,10 @@ robot-api (0) unstable; urgency=low
   * (No changelog)
 
  -- Andrew Barrett-Sprot <abarrettsprot@gmail.com>  Sun, 19 Nov 2017 10:47:00 +0000
+
+
+robot-api (2018.2.0) unstable; urgency=low
```
PeterJCLaw: For some reason I was expecting the new version to be at the top. I'm not sure it matters which way around they go, but have we checked that this builds the debian package as we'd want?

Adimote: ah looks like you might be right here

Adimote: fixed in

PR #88, opened 2018-03-29
rename socket.._single_retry to be more apt

Adimote:
This has been on my TODO list for 4 months.

remove the name of `_socket_with_single_retry` (because it actually retries more than once)
to be replaced by `socket_with_backoffs` (because it uses backoffs).


PeterJCLaw:
@Adimote thoughts on #78 (which renders this somewhat redundant)?

PR #89, opened 2018-04-07
Stop robots moving at the end of matches

antonnikitin97:


robot/game.py
```diff
@@ -1,10 +1,37 @@
+import logging
+import time
 from enum import Enum
+from threading import Thread
 from typing import NewType
 
+import _thread
```
trickeydan: Why isn't this up with the rest of the imports?

robot/game.py
```diff
@@ -1,10 +1,37 @@
...
 
+LOGGER = logging.getLogger(__name__)
+
+
+def kill_after_delay(timeout_seconds, exit_message):
+    """
+    Interrupts main process after the given delay.
+    """
+
+    end = time.time() + timeout_seconds
```
trickeydan: I don't think that `end` is a particularly great variable name here.

Perhaps `end_time`?

robot/game.py
```diff
@@ -1,10 +1,37 @@
...
+
+    def worker():
+        while time.time() < end:
+            time.sleep(0.1)
+
+        LOGGER.info("Timeout %r expired: %s", timeout_seconds, exit_message)
```
trickeydan: This currently would display as:
`Timeout 150 expired: Game Over!`

I think this could be a bit confusing to people reading through the logs as it doesn't say what the number means.

`Timeout 150s expired: Game Over!`

robot/game.py
```diff
@@ -1,10 +1,37 @@
...
+
+        # interrupt the main thread to close the user code
```
trickeydan: /s/interrupt/Interrupt

/s/close/kill

robot/power.py
```diff
@@ -60,6 +62,9 @@ def wait_start(self):
         self.set_start_led(False)
         LOGGER.info("Starting user code.")
 
+        if GameState.mode == GameMode.COMPETITION:
```
trickeydan: Can you move this `if` before line 63 `LOGGER.info("Starting user code.")`

robot/game.py
```diff
@@ -1,10 +1,9 @@
 import logging
 import time
+import _thread
 from enum import Enum
```
trickeydan: Please add a line break here to separate out the types of imports.
e.g
```
import module1
import module2

from module10 import AClass
from ...

File: tests/test_power.py [hunk 7]
```diff
@@ -119,3 +140,5 @@ def test_fractional_buzz_duration(self):
 
     def tearDown(self):
         self.mock.stop()
+
+from robot.power import PowerBoard
```
RealOrangeOne: Why is this import all the way down here?

robot/power.py
```diff
@@ -20,6 +21,15 @@ class PowerBoard(Board):
...
     }
 
+    def __init__(
+        self,
+        *args,
+        on_start_signal: Callable[[], None]=lambda: None,
+        **kwargs
+    ) -> None:
+        super().__init__(*args, **kwargs)
+        self._on_start_signal = on_start_signal
```
RealOrangeOne: Doesn't look like this is ever called (PR possibly incomplete?)

PeterJCLaw:
@RealOrangeOne, @trickeydan please could you take another look at this? I've fixed up the tests.

robot/game_specific.py
```diff
@@ -6,6 +6,9 @@
 WALL = set(range(0, 28))  # 0 - 27
 
 # Currently for SB2018
+GAME_DURATION_SECONDS = 150
+GAME_EXIT_MESSAGE = "Game over!"
```
RealOrangeOne: The exit message isn't game specific. This should probably just be baked into the log message in `game.py`

trickeydan: The reason for that was to make the function as general as possible, decreasing coupling. Also, we cannot guarantee that this will not change in the future.

RealOrangeOne: Agreed, but this file isnt a general constants file, it's for game-specific things, which this message isnt. It's also ok to change things in a string without abstracting it to a variable.

trickeydan: @PeterJCLaw I recall you suggested that `GAME_EXIT_MESSAGE` was a constant. Do you remember the justification for this?

antonnikitin97: @trickeydan I believe the justification was to make it as general as possible. (We may want to change it year on year)

trickeydan:
@PeterJCLaw I have just tested this on a pi and the code did not die. Will investigate later after a meeting I have to attend

trickeydan:
It is not clear what is going wrong with it

PeterJCLaw:
Hrm, that's a shame. @trickeydan do you have the logs from your test? I'd be interested to see whether they show that the main process is being killed in the way we think it should be or not. (I'm expecting this will tell us whether the issue is in the new mechanism here or whether we've found an issue in the `robotd` interactions)

Adimote:
from memory I believe he said it didn't show anything on the logs

trickeydan:
I was unable to get any logs beyond the robot-api output.

PeterJCLaw:
Ah, I was mainly thinking about the `log.txt` on the memory stick.

If we could get the `robotd`` logs that might also be handy. They'll be in the systemd journal: `sudo journalctl -u robotd`.
The `runusb` log might also be useful: `sudo journalctl -u runusb`

PeterJCLaw:
@trickeydan please could you also share the `main.py` that you were testing with?

I suggest we test with a `main.py` something like this:
``` python
import time
from robot import Robot
r = Robot()
while True:
    print(time.time())
    time.sleep(0.1)
```

This will give us useful log output to inspect while ensuring that the `main.py` doesn't exit of its own accord before we time it out.

We might also want to tweak the timeout to be a bit shorter in development.

PeterJCLaw:
@trickeydan I've just tested this with the following script:
``` python
import time
import robot

r = robot.Robot()

print(r.mode, r.zone)

while True:
    print(time.time())
    time.sleep(0.1)
```
It seems to work as expected for me. The gotcha is that you need to be in competition mode (i.e: have a zone file around) for the timeout to be configured.

I've added logging of the current mode & zone which should help with this.

RealOrangeOne:
This has been tested on a fresh `pi-image` build and works fine!

robot/game_specific.py
PeterJCLaw: At the time I was envisioning that the helper was a fully generic alarm mechanism (this was before @prophile mentioned the existence of [`signal.alarm`](https://docs.python.org/3.5/library/signal.html#signal.alarm)).

I don't feel strongly about this constant (though, obviously, the duration does belong in `game_specific.py`).

RealOrangeOne: I agree. The duration should be in `game_specific.py`, however I don't think the message should be. Perhaps baking it into the log would be better

PeterJCLaw: 588762b

PR #91, opened 2018-04-14
Log the current zone and mode on startup

RealOrangeOne:
So it's more obvious what zone a robot is running in, which may be useful for debugging

### DIFF
--- a/robot/robot.py
+++ b/robot/robot.py
@@ -56,6 +56,10 @@ class Robot:
         LOGGER.info("Robot (v{}) Initialising...".format(__VERSION__))
         self._assert_has_power_board()
         self.power_board.power_on()
+        LOGGER.info("Zone: {zone}. Mode: {mode}".format(
+            zone=self.zone,
+            mode=self.mode
+        ))
 
         if wait_for_start_button:
             self.power_board.wait_start()


PeterJCLaw:
@RealOrangeOne this [already happens](https://github.com/sourcebots/robot-api/blob/master/robot/robot.py#L113-L118) when the start button is pressed, do we need it on startup as well?

Alternatively, should we remove that duplication?

RealOrangeOne:
Ah, didn't realise that existed. Perhaps we remove it from the start signal? Not sure it's necessary there?

PeterJCLaw:
I put it in the start signal primarily as a debugging aid for testing the start signal behaviour -- since it uses the same instance of the game state it's reliable as to whether the game timeout would get started. I'd like to leave that in place, even if we dialed it down to being `debug` rather than `info`.
However: once we have that, I'm not sure what the benefits to having it elsewhere are?

RealOrangeOne:
Personally I don't see why having it in the start signal is better, but as long as the values are presented, that's fine with me!

PeterJCLaw:
@RealOrangeOne for clarity:

Given that the zone number and mode can change (by, for example, unplugging the USB stick and plugging in a different one, while the robot is running):
 - from the competitors perspective I agree that it doesn't matter when we log it, since the one logged is merely a single sample
 - from a debugging-the-kill-code perspective, there's a value in it being in the start signal handler so that we can have confidence that the value logged is the value the code saw at the point it made the decision about whether to setup the kill code.
   I had originally thought that by loading the game state into a local variable at that point I would guarantee that consistency, however as it doesn't work like that (instead the `mode` and `zone` properties query RobotD each time) this isn't guaranteed.

I still think from a debugging perspective the start signal is the right time to do this as it reduces two race-condition windows:
 - change between logging and kill-code setup
 - plugging the USB sticks in in the wrong order

PR #93, opened 2018-06-04
Deprecated is_wall_marker/is_token_marker methods

lockwooda:
https://trello.com/c/tHwRNdfo/63-deprecate-remove-iswallmarker-istokenmarker-methods
As agreed, we are deprecating is_wall_marker and is_token_marker and encouraging the use of marker.id instead of these two methods.

Docs removed in https://github.com/sourcebots/docs/pull/139

### DIFF
--- a/robot/markers.py
+++ b/robot/markers.py
@@ -123,14 +123,6 @@ class Marker:
         return self.spherical.distance_metres
 
     # Helper functions, Might need to vary these per-game
-    def is_wall_marker(self) -> bool:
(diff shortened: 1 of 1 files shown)


robot/markers.py
```diff
@@ -123,14 +121,6 @@ def distance_metres(self) -> Metres:
         return self.spherical.distance_metres
 
     # Helper functions, Might need to vary these per-game
```
PeterJCLaw: I think this comment line can be removed as well.

PR #97, opened 2018-08-07
Add support for controlling separate power outputs on power board

RealOrangeOne:
Implements support for controlling separate power outputs, exposed by https://github.com/sourcebots/robotd/pull/51

### DIFF
--- a/robot/power.py
+++ b/robot/power.py
@@ -1,7 +1,8 @@
+import enum
 import logging
 import time
 from typing import Callable
-import enum
+
 from robot.board import Board
 
 LOGGER = logging.getLogger(__name__)
@@ -56,8 +57,11 @@ class PowerBoard(Board):
         self._send_and_receive({'power': False})
 
     def set_output(self, output: PowerOutput, value: bool) -> None:
-        self.send_and_receive({
-            'power-output': output.value
+        """
+        Turn off individual power output.
+        """
+        self._send_and_receive({
+            'power-output': output.value,
             'power-level': value,
         })
 
--- a/robot/power.py
+++ b/robot/power.py
@@ -58,7 +58,7 @@ class PowerBoard(Board):
 
     def set_output(self, output: PowerOutput, value: bool) -> None:
         """
-        Turn off individual power output.
+        Turn on and off individual power outputs.
         """
         self._send_and_receive({
             'power-output': output.value,
--- a/robot/power.py
+++ b/robot/power.py
@@ -1,10 +1,22 @@
 import logging
 import time
 from typing import Callable
-
+import enum
 from robot.board import Board
 
 LOGGER = logging.getLogger(__name__)
+
+
+# Keep this in sync with `robotd`
+class PowerOutput(enum.Enum):
+    """An enumeration of the outputs on the power board."""
+
+    HIGH_POWER_1 = 0
+    HIGH_POWER_2 = 1
+    LOW_POWER_1 = 2
+    LOW_POWER_2 = 3
+    LOW_POWER_3 = 4
+    LOW_POWER_4 = 5
 
 
 class PowerBoard(Board):
@@ -42,6 +54,12 @@ class PowerBoard(Board):
         Turn off power to all power board outputs.
         """
         self._send_and_receive({'power': False})
+
+    def set_output(self, output: PowerOutput, value: bool) -> None:
+        self.send_and_receive({
+            'power-output': output.value
+            'power-level': value,
+        })
 
     def set_start_led(self, value: bool):
         """Set the state of the start LED."""
--- a/robot/power.py
+++ b/robot/power.py
@@ -12,12 +12,12 @@ LOGGER = logging.getLogger(__name__)
 class PowerOutput(enum.Enum):
     """An enumeration of the outputs on the power board."""
 
-    HIGH_POWER_0 = 0
-    HIGH_POWER_1 = 1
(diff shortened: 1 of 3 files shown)


robot/power.py
```diff
@@ -43,6 +56,15 @@ def power_off(self):
         """
         self._send_and_receive({'power': False})
 
+    def set_output(self, output: PowerOutput, value: bool) -> None:
+        """
+        Turn off individual power output.
```
trickeydan: What is our standard for docstrings? I see two different styles here

tests/test_power.py
```diff
@@ -139,3 +139,12 @@ def test_fractional_buzz_duration(self):
             'duration': 500,
         })
         self.assertIsInstance(msg['buzz']['duration'], int)
+
+    def test_separate_power(self):
+        self.power_board.clear_queue()
+        self.robot.power_board.set_output(PowerOutput.HIGH_POWER_1, True)
```
trickeydan: Any chance of looping through all the enums in this test?

robot/power.py
trickeydan: Looks like we are inconsistent **everywhere**. This should probably be addressed in a separate PR

PeterJCLaw: This does more than just turning them off -- you can turn them _on_ too.

Also Maybe add "an"?

"Turn on or off an individual power output".

robot/power.py
```diff
@@ -7,6 +8,18 @@
 LOGGER = logging.getLogger(__name__)
 
 
+# Keep this in sync with `robotd`
+class PowerOutput(enum.Enum):
+    """An enumeration of the outputs on the power board."""
```
PeterJCLaw: We should validate that the name -> id mapping here is correct (I'm not sure the best way to do that other than getting a power board and trying it).

tests/test_power.py
PeterJCLaw: Not sure what that would actually usefully achieve?

robot/power.py
PeterJCLaw: Does the linter (`flake8-docstrings`) not tell you whether this is the expected format?

PeterJCLaw: Ah, it seems not; I'll add details on #98.

robot/power.py
```diff
@@ -43,6 +56,15 @@ def power_off(self):
...
         self._send_and_receive({'power': False})
 
+    def set_output(self, output: PowerOutput, value: bool) -> None:
+        """
+        Turn off individual power output.
+        """
+        self._send_and_receive({
+            'power-output': output.value,
+            'power-level': value,
+        })
```
prophile: Can we typecheck these? Passing in 0.5 as the power level to get 50% power is the type of thing which teams would attempt and that'll produce a very odd error message at the mo.

robot/power.py
```diff
@@ -7,6 +8,18 @@
 LOGGER = logging.getLogger(__name__)
 
 
+# Keep this in sync with `robotd`
+class PowerOutput(enum.Enum):
+    """An enumeration of the outputs on the power board."""
+
+    HIGH_POWER_0 = 0
```
PeterJCLaw: We should probably match the names in the docs -- either `H1` or `HIGH_CURRENT_1`.

robot/power.py
```diff
@@ -43,6 +56,17 @@ def power_off(self):
         """
         self._send_and_receive({'power': False})
 
+    def set_output(self, output: PowerOutput, value: bool) -> None:
```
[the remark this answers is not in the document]
PeterJCLaw: Hrm, I was looking at the GPIOs here, though the power board is all methods, so maybe this is better anyway.

PeterJCLaw: Hrm on writing the docs I'm realising that this API is actually quite unclear as it stands. It's non-obvious to me what it is that:
```python
r.power_board.set_output(PowerOutput.HIGH_POWER_1, True)
```
actually does. I think it's the combination of `output` and `True`.

Perhaps something like this might be clearer:
```python
r.power_board.disable_output(PowerOutput.HIGH_POWER_1)
r.power_board.enable_output(PowerOutput.HIGH_POWER_1)
```

I'll put up a PR for that.

PeterJCLaw: In fact, `power_on_output(...)` and `power_off_output(...)` is better as it follows the existing naming.

PeterJCLaw: this is now https://github.com/sourcebots/robot-api/pull/101

robot/power.py
```diff
@@ -7,6 +8,18 @@
 LOGGER = logging.getLogger(__name__)
 
 
+# Keep this in sync with `robotd`
+class PowerOutput(enum.Enum):
```
PeterJCLaw: We need to expose this as part of the API -- we'll need to import it into `__init__.py` and re-export it from there.

robot/power.py
trickeydan: We should match the labelling on the boards surely...?

PeterJCLaw: I'm not sure what the boards say, though I think they use `H1`. Do you have a power board you could check with @trickeydan?

robot/power.py
PeterJCLaw: a6f08ea

robot/power.py
PeterJCLaw: (this was checked by @RealOrangeOne)

PR #99, opened 2018-08-07
Update the game length for Smallpeice 2018

trickeydan:


robot/game_specific.py
```diff
@@ -6,7 +6,7 @@
 WALL = set(range(0, 28))  # 0 - 27
 
 # Currently for SB2018
```
prophile: Maybe update to Smallpeice 2018 :)

